PR #18, opened 2018-02-12
docs: add promo buttons

miralemd:
Added buttons to "Get started" and "Examples"


carlhannes:
Just need to check if this works with the new static-scrolling aswell..

PR #22, opened 2018-02-13
docs: add og-image

miralemd:
add logo as og:image


carlhannes:
It’s actually a good idea if this image is not transparent, so something like the twitter bio image is probably better

PR #43, opened 2018-03-02
 If Horizontal Axis dock is left and text is aligned left, align axis line right

yianni-ververis:
**Checklist**

- [ ] tests added
- [x] documentation updated


yianni-ververis:
Live example
https://webapps.qlik.com/qdt-components/plain-html/picasso-horizontalbarchart.html

cbt1:
Hi Yianni, thanks for the PR.

We're looking into a way of adding a bit more flexibility into the labels on the axis, for example by adding a property to align the labels along the cross-axis (ex. x-axis on a vertically docked axis). Which we think would solve your issue and provide a bit more functionality, as it would also allow labels to be centered.

You're more then welcome to take a look at an issue I've created on this #48 .

A possible workaround for your issue, is to have 2 axis component. One that renders the labels and another that render only the line.

yianni-ververis:
interesting, so justify will do in a vertical axis? Did you bump up the version?

cbt1:
It would allow you to align the labels to the left, such that they start on the left side, while still having the line on the side closes to the center area. Is that correct understanding of what your're looking for?

No we have not bumped the version.

yianni-ververis:
here is the chart,  I will try your solution in a bit after I merge everything

https://webapps.qlik.com/qdt-components/plain-html/picasso-horizontalbarchart.html

miralemd:
@yianni-ververis closing this for now, please open new PR per feature

PR #51, opened 2018-03-08
docs: Minor spelling fix

carlioth:
Found minor spelling issues when reading through the doc.


carlhannes:
Awesome! Could you run `npm run docs` and commit the results of that aswell, so the `dist/` gets updated? @miralemd I assume we need to re-publish the website after doing this?

PR #54, opened 2018-03-16
chore: Simplify rollup

carlhannes:
## What has been done?
* Sourcemaps are now always built
* Only one file is outputted instead of two, `dist/picasso.js` (this is minifed on build, but still mapped to the sourcemap)
* Introduced rollup-plugin-clean for `dist` folder so the .min. file does not keep laying around
* On watch it skips uglifying to build faster
* `npm run build:dev` removed

## Checklist
- [ ] tests added
- [ ] documentation updated

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,4 @@
-dist/*
+**/dist/*
 test/unit/coverage/*
 test/component/coverage/*
 docs/node_modules/*
--- a/package-lock.json
+++ b/package-lock.json
@@ -9530,6 +9530,15 @@
         "rollup-pluginutils": "1.5.2"
       }
     },
+    "rollup-plugin-clean": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/rollup-plugin-clean/-/rollup-plugin-clean-1.0.0.tgz",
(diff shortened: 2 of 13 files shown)


carlhannes:
```
X FAILED: Brushing tap ( 277ms )
     unknown error: Cannot read property 'findShapes' of undefined
  (Session info: chrome=64.0.3282.186)
  (Driver info: chromedriver=2.34.522913 (36222509aa6e819815938cbf2709b4849735537c),platform=Linux 4.4.0-109-generic x86_64)
     /home/circleci/picassojs/packages/picasso.js/test/integration/chart/brushing/brushing.int.js
```
@cbt1 ?

cbt1:
The integration tests are using the minified file

carlhannes:
:| why doesn’t it use ”Main”?


mån 19 mars 2018 kl. 06:51 skrev Christofer Bengtsson <
notifications@github.com>:

miralemd:
Since `.min.js` has been part of the bundle so far, this would be a BREAKING CHANGE. I would prefer we keep `.min.js` (just a copy of `.js`) until we reach `v1.0.0`

PR #68, opened 2018-04-23
feat: share per slice

cbt1:
Adds a share value (0-1) on each slice. Also added a filter to remove any data value that is negative or NaN, previously all those values where passed to the renderer but didn't have any size/area.

This change exposes a way to show share percentage in for example labels and tooltip.

**Checklist**

- [x] tests added
- [ ] documentation updated

### DIFF
--- a/packages/picasso.js/test/unit/core/chart-components/pie.spec.js
+++ b/packages/picasso.js/test/unit/core/chart-components/pie.spec.js
@@ -44,8 +44,38 @@ describe('pie', () => {
       stroke: '#ccc',
       strokeWidth: 1,
       opacity: 1,
-      data: { value: 1, label: '1' }
+      data: { value: 1, label: '1' },
+      desc: {
+        share: 1
+      }
     });
+  });
+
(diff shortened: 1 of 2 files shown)


packages/picasso.js/test/unit/core/chart-components/pie.spec.js
```diff
@@ -44,18 +44,48 @@ describe('pie', () => {
...
+  });
+
   describe('arcValue', () => {
     it('should default to data.value', () => {
       expect(arcValue({}, {
         data: { value: 4 }
       })).to.equal(4);
     });
 
-    it('should pioritize .arc', () => {
```
carlhannes: this isn't a pun then? Pie-oritize?

carlhannes: or more likely pi-oritize...

cbt1: lol

Issue #70, opened 2018-05-07
The graph injection fail silently, no error.


SuperstrongBE:
### Description
First use of picasso.js in a ES6 project. Code are copy pasted from the github project page.

[Description of the issue]
The graph injection fail silently.

### Steps to Reproduce
With webpack, picasso installed via NPM. 
1. [Step one]
import picasso as dependency 
2. [Step two]
copy paste code from https://github.com/qlik-oss/picasso.js

##### Expected behavior
The plot graph rendered in the element specified in the definition.

##### Actual behavior
Graph is not rendered and dom element specified in the definition  remain empty (id checked more then twice :) ). 
Created and mounted callback are fired. Mounted dom element specified params is the expected dom element specified in the definition. No error in console.

### Versions

* picasso.js: 
 "picasso.js": "^0.7.0"
* Browser: 
Chrome Version 66.0.3359.139 (Build officiel) (64 bits)
Firefox 59.0.2 (64 bits)
* [Other relevant versions]

miralemd:
@SuperstrongBE Thanks for bringing this up, the example is updated - could you check if it works for you now?

PR #73, opened 2018-05-13
docs: update example in README

miralemd:
Example in readme wasn't updated when we switched function callbacks, likely cause to #70 .


stoffeastrom:
Yes, looks good! Tried it on [stackblitz](https://js-v8rzpk.stackblitz.io/). As now stated in the `README` make sure the container has a width and height!

PR #76, opened 2018-05-15
chore: update tooling

stoffeastrom:


stoffeastrom:
## Major tooling update

All dependencies updated to **latest**
Use `hoisting`
Fix all linting after update
Fix all unit tests after update
Fix flaky integration tests and start using `async` `await`

## Developing

After `hoisting` you have to execute your commands from the root. To run a specific package do:

`npm test -- --scope picasso.js`
`npm run build -- --scope picasso.js`
...

PR #82, opened 2018-05-17
style: fix lint error from #76

T-Wizard:
from update of eslint?


T-Wizard:
no error (missed npm install)

PR #92, opened 2018-06-04
feature: preact as dom renderer

nilzona:
- Update eslint config to enable jsx in javascript
- Replace snabbdom in favour for preact.
- Switch from es2015 babel preset to env (recommended)

`rollup-plugin-commonjs` > 8.2.3 and `babel-plugin-transform-react-jsx` are not compatible. Hence locking version to 8.2.3 until issue (https://github.com/rollup/rollup-plugin-commonjs/issues/247) is resolved


cbt1:
This seems like a breaking change. What are the differences between snabbdom and preact API wise? I would suggest that you create a new preact-renderer instead of replacing the dom-renderer.

nilzona:
@cbt1  The idea was to not have two dom-renderers. snabbdom and preact exposes the same hyperscript api `h`. snabbdom was not directly exposed but wrapped in `dom-renderer.js`. It's still the same api but it's wrapping preact instead so that we can start writing jsx instead of hyperscript.

cbt1:
@nilzona Ok, I've yet to find anything similar to snabbdom "hooks" in the preat API ref. for the `h` function. Which in my eyes would mean a breaking change.

nilzona:
@cbt1  Yep, you're right, that's a breaking change. It is very unlikely that it has any impact since we haven't documented anything about the internals of the functions that dom-renderer is providing.
I still don't think we should have two dom-renderers.

nilzona:
Waiting with this for a while until we refactor the code that is built on snabbdom's hooks api

PR #100, opened 2018-06-11
feat: override range

nguyenlam1:
Provide the option to override range values from the explicit values instead of always injecting them.


packages/picasso.js/src/core/scales/color/categorical.js
```diff
@@ -12,6 +12,24 @@ const DEFAULT_EXPLICIT_SETTINGS = {
   range: []
 };
 
+const overrideRange = (range, domain, explicit) => {
+  const overridenRange = range.slice();
```
cbt1: Would prefer a name not so similar to the function name. Ex: "newRange", instead of something that is more or less the same.

packages/picasso.js/src/core/scales/color/categorical.js
```diff
@@ -12,6 +12,35 @@ const DEFAULT_EXPLICIT_SETTINGS = {
...
 };
 
+/**
+ * Overrides the range with the values specified in the explicit object if possible.
+ * If there are no overrides, the values from explicit object will be injected into
+ * the range in the order specified by the domain.
+ * If the index is out of bounds, no injection or override will be performed.
+ * @param {Array} range - colors
+ * @param {Array} domain - values to be colored
+ * @param {Object} explicit - Object containing explicit domain and range to be overriden or injected.
```
miralemd: add `@ignore`

packages/picasso.js/src/core/scales/color/categorical.js
nguyenlam1: fixed!

packages/picasso.js/src/core/scales/color/categorical.js
nguyenlam1: fixed!

carlhannes: Sure you've added it? Can't see it @nguyenlam1

nguyenlam1:
@carlhannes Thanks! :)

packages/picasso.js/src/core/scales/color/categorical.js
```diff
@@ -36,8 +66,10 @@ export default function scaleCategorical(settings = {}, data = {}, resources = {
   const theme = resources.theme;
   const stgns = resolveSettings(settings, DEFAULT_SETTINGS, { data, resources });
   stgns.explicit = resolveSettings(settings.explicit, DEFAULT_EXPLICIT_SETTINGS, { data, resources });
+  const haveOverride = settings.explicit && settings.explicit.override;
```
miralemd: add `override: false` to `DEFAULT_EXPLICIT_SETTINGS` and use `stngs.explicit.override`

PR #146, opened 2018-08-15
feat: Enable selectPivotCells for brush

nguyenlam1:
**Checklist**

- [x] tests added
- [x] documentation updated

Enables brush to detect a selectPivotCells(https://help.qlik.com/en-US/sense-developer/June2018/APIs/EngineAPI/services-GenericObject-SelectPivotCells.html) call by providing the layout as third parameter.


plugins/q/test/unit/q-brush.spec.js
```diff
@@ -199,6 +199,130 @@ describe('q-brush', () => {
     });
   });
 
+  describe('selectPivotCells', () => {
+    const layout = {
+      qHyperCube: {
```
miralemd: should move this into `beforeEach` to ensure pure object for each test

nguyenlam1: Fixed!

PR #147, opened 2018-08-16
feat: tooltip

cbt1:
A customizable tooltip component that can display complementary information. Typically used when moving the mouse curser over a node or selecting a node.

tooltip

## Example

```js
picasso.chart({
  settings: {
    interactions: [{
      type: 'native',
      events: {
        mousemove(e) {
          const tooltip = this.chart.component('<tooltip-key>');
          tooltip.emit('show', e);
        },
        mouseleave(e) {
          const tooltip = this.chart.component('<tooltip-key>');
          tooltip.emit('hide');
        }
      }
    }],
    components: [
      {
        key: '<tooltip-key>',
        type: 'tooltip'
      }
    ]
  },
  ...
});
```

## Events

The tooltip component is a reactive component that depend on events to render something.

### `show` event

The event takes two parameters, the first is the pointer event and the second is an optional object which may contain specific nodes to use when rendering the tooltip. If the second parameter is omitted, the pointer event will be used to perform a node lookup on the chart instance.

With a single event parameter:
```js
picassoInstance.component('<key value of tooltip>').emit('show', event);
```

With optional parameters:
```js
const nodes = picassoInstance.shapesAt({ x: 1, y: 2, width: 3, height: 4}); // An array of nodes
picassoInstance.component('<key value of tooltip>').emit('show', event, {
  nodes,
  delay: 50, // Optional, the component setting value is used as default
  duration: 2000 // Optional, the component setting value is used as default
}); // The event is dispatched and if no new events are triggered until the delay has passed, a tooltip is displayed
```

### `hide` event

Hide the tooltip. Will hide the tooltip if it's currently is displayed, if not the event is ignored.

```js
picassoInstance.component('<key value of tooltip>').emit('hide');
```

## Extracting data and generating custom content

As the main purpose of a tooltip is often to displayed complementary data when hovering or selecting a node, controlling which data to displayed is crucial. Luckily there are two settings available to ensure that correct data is displayed in the tooltip.

### Extracting data

The `extract` setting is executed for each node and is expected to return a data representation. By default the extracted data is a string with the value of the node. But could be any data structure, such as an object or an array of objects.

#### Default

```js
const extract = ({ node }) => node.data.value;
```

### Generating content

The `content` setting is responsible for generating virtual nodes using the HyperScript API and is expected to return an array of virtual nodes.

#### Default

```js
{
  settings: {
    content: ({ h, data }) => data.map(datum => h('div', {}, datum))
  }
}
```

### Example - Formatting values

Here we format the value using a formatter and return the output as extracted data.
```js
{
  settings: {
    extract: ({ node, resources }) => resources.formatter('<name-of-a-formatter>')(node.data.value)
  }
}
```

But sometimes we want to give the value some context, for example by labeling it. As such we change the data structure to an  object instead to make it more clear what the data represent. As we do, the default `content` generator function will complain because it doesn't understand our new data structure, so we must also update the `content` setting.
```js
{
  settings: {
    extract: ({ node, resources }) => ({
      label: node.data.label,
      value: resources.formatter('<name-of-a-formatter>')(node.data.value)
    }),
    // We go from a basic content generator, to one that understand our new data structure
    content: ({ h, data }) => data.map(datum => h('div', {}, `${datum.label}: ${datum.value}`))
  }
}
```

## Placement strategies

The placement strategy defines how the tooltip is positioned in relation to a set of nodes or a node. The built-in strategies all have a auto docking feature, such that if the tooltip doesn't fit inside its designated area, the docking position may be adjusted to a position where it does fit. Note that there may still be scenarioes where the tooltip doesn't fit, for example when the designated area just cannot fit the tooltip at all.

There are 3 available strategies:
* `pointer` - is the default and use the current pointer position to place the tooltip.
* `bounds` - use the bounding rectangle of the first node. Note the depending on the type of shape, the position of the tooltip may indicate that it's not over the visual area of the shape.
* `slice` - is built to work together with the `pie` component, such that the tooltip is placed at the arc of a slice.

### Example

```js
// Define as a string
{
  settings: {
    placement: 'pointer'
  }
}

// Or as an object
{
  settings: {
    placement: {
      type: 'pointer',
      dock: 'top',
      offset: 8,
      area: 'target'
    }
  }
}

// Or as a function
{
  settings: {
    placement: () => ({
      type: 'pointer',
      dock: 'top',
      offset: 8,
      area: 'target'
    })
  }
}
```

### Example - Custom placement function

When the built in placement strategies are not sufficient, it possible to create your own placement function.

```js
{
  settings: {
    placement: {
      fn: () => {
        // Do some magic here
        return {
          {
            computedTooltipStyle: {
              left: '10px',
              top: '20px',
              color: 'red'
            },
            computedArrowStyle: {
              left: '0px',
              top: '0px',
              color: 'red'
            },
            dock: 'top'
          }
        };
      }
    }
  }
}
```

## API Reference

```js
{
  /**
   * How long the tooltip is visible, in milliseconds
   * @type {number=}
   */
  duration: 8000,
  /**
   * Delay before the tooltip is rendered, in milliseconds
   * @type {number=}
   */
  delay: 500,
  /**
   * Reduce incoming nodes to only a set of applicable nodes. Is called as a part of the `show` event.
   * @type {function=}
   * @returns {array} An array of nodes
   */
  filter: nodes => nodes.filter(node => node.data && typeof node.data.value !== 'undefined'),
  /**
   * Extract data from a node.
   * @type {function=}
   * @returns {object} An array of data
   */
  extract: ctx => ctx.node.data.value,
  /**
   * Content generator. Extracted data is available in the `data` property, where each value in the area is the extracted datum from a node.
   * @type {function=}
   * @returns {object[]} Array of h objects
   */
  content: ({ h, data }) => data.map(datum => h('div', {}, datum)),
  /**
   * Comparison function. If evaluted to true, the incoming nodes in the `show` event are ignored. If evaluated to false, any active tooltip is cleared and a new tooltip is queued.
   *
   * The function gets two parameters, the first is the currently active set of nodes, if any, and the second is the incoming set of nodes. By default the two set of nodes are considered equal if their data attributes are the same.
   * @type {function=}
   * @returns {boolean}
   */
  isEqual: (prev, curr) => prev.length &&
    prev.length === curr.length &&
    prev.every((p, i) => curr[i] && JSON.stringify(p.data) === JSON.stringify(curr[i].data)),
  /**
   * @typedef {object=}
   */
  placement: {
    /**
     * Available types: [pointer | bounds | slice]
     * @type {string=}
     */
    type: 'pointer',
    /**
     * Docking position of the tooltip. Available positions: [left | right | top | bottom | auto]
     * @type {string=}
     */
    dock: 'auto',
    /**
     * Distance from the content area to the tooltip position, in px.
     * @type {number=}
     */
    offset: 8,
    /**
     * Specify the limiting area, where target is the component area unless the appendTo property is set, in which case it referes to the appendTo element. Viewport is the browser viewport.
     *
     * Available options are: [viewport | target]
     * @type {number=}
     */
    area: 'viewport'
  },
  /**
   * Set tooltip class.
  * @type {object<string, boolean>=}
  */
  tooltipClass: {},
  /**
   * Set content class.
  * @type {object<string, boolean>=}
  */
  contentClass: {},
  /**
   * Set arrow class.
  * @type {object<string, boolean>=}
  */
  arrowClass: {},
  /**
   * Content direction [ltr | rtl]
   * @type {string=}
   */
  direction: 'ltr',
  /**
   * Explicitly set a target element. This allows the tooltip to attach itself outside the picasso container.
   * @type {HTMLElement=}
   */
  appendTo: undefined,
  /**
   * Component lifecycle hook. Called before the tooltip is displayed.
   * @type {function=}
   */
  beforeShow: undefined,
  /**
   * Component lifecycle hook. Called after the tooltip have been displayed.
   * @type {function=}
   */
  afterShow: undefined,
  /**
   * Component lifecycle hook. Called before the tooltip is hidden.
   * @type {function=}
   */
  beforeHide: undefined,
  /**
   * Component lifecycle hook. Called when the toolip is hidden. By default this deletes the tooltip element.
   * @type {function=}
   */
  onHide: undefined,
  /**
   * Component lifecycle hook. Called after the tooltip is hidden.
   * @type {function=}
   */
  afterHide: undefined
}
```

### Style

```js
{
  tooltip: {},
  content: {
    backgroundColor: '$gray-25',
    color: '$font-color--inverted',
    fontFamily: '$font-family',
    fontSize: '$font-size',
    lineHeight: '$line-height',
    borderRadius: '4px',
    padding: '8px',
    opacity: 0.9
  },
  arrow: {
    position: 'absolute',
    width: '0px',
    height: '0px',
    borderStyle: 'solid',
    color: '$gray-25',
    opacity: 0.9
  },
  'arrow-bottom': {
    borderTopColor: 'transparent',
    borderLeftColor: 'transparent',
    borderRightColor: 'transparent'
  },
  'arrow-top': {
    borderBottomColor: 'transparent',
    borderLeftColor: 'transparent',
    borderRightColor: 'transparent'
  },
  'arrow-right': {
    borderTopColor: 'transparent',
    borderLeftColor: 'transparent',
    borderBottomColor: 'transparent'
  },
  'arrow-left': {
    borderTopColor: 'transparent',
    borderBottomColor: 'transparent',
    borderRightColor: 'transparent'
  }
}
```


packages/picasso.js/src/web/components/tooltip/extractor.js
```diff
@@ -0,0 +1,28 @@
...
+      scale: chart.scale,
+      formatter: chart.formatter
+    },
+    scale,
+    h
+  };
+
+  const data = [];
+  nodes.forEach((node) => {
+    if (typeof props.extract === 'function') {
```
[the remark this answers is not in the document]
cbt1: @nguyenlam1 It must be a function. If not defined in settings than a default setting is used. Per default the extracted data is the value of the node.

packages/picasso.js/src/web/components/tooltip/placement.js
```diff
@@ -0,0 +1,416 @@
...
+    y: dy + componentBounds.y + (componentBounds.height / 2)
+  };
+
+  const {
+    start,
+    end,
+    outerRadius
+  } = node.desc.slice;
+
+  // origin is at 12 o clock, clockwise, so it is transform to origin at 3 a clock
```
nguyenlam1: Could you add a comment that it starts counting clockwise at Pi / 2 and that these calculations start before the transform? Would be greatly appreciated for me :)

packages/picasso.js/src/web/components/tooltip/render.js
```diff
@@ -0,0 +1,62 @@
+import extend from 'extend';
+
+function resolveClasses(props, opts) {
+  return {
+    tooltip: typeof props.tooltipClass === 'function' ? props.tooltipClass({ dock: opts.dock }) : props.tooltipClass,
```
nguyenlam1: So, tooltipClass, contentClass and arrowClass can be a function as well? Shouldn't that be specified in the API reference?

packages/picasso.js/src/web/components/tooltip/placement.js
cbt1: Well it sort of does state that. By I'll try and make it more clear

packages/picasso.js/src/web/components/tooltip/render.js
cbt1: Most, if not all, settings can be a function, but to keep the API clean. I prefer to specify only the required type and instead use the documentation to add examples where you use a function. That also adds some flexibility when it comes to breaking changes.

PR #148, opened 2018-08-16
feat: stroke-linejoin 

carlhannes:
Introducing stroke-linejoin property for display objects & box, line and pie components, including tests.

Also default to round for pie to prevent "sharp" behaviour on slices

**Checklist**

- [x] tests added
- [x] documentation updated


cbt1:
I would prefer if you use the correct kebab-case to camelCase convertion for `stroke-linejoin` => `strokeLinejoin`

docs/spec.json
```diff
@@ -1047,6 +1047,11 @@
                   "defaultValue": 1,
                   "type": "number"
                 },
+                "strokeLineJoin": {
+                  "optional": true,
+                  "defaultValue": "'miter'",
+                  "type": "number"
```
nguyenlam1: Shouldn't the type be string here?

carlhannes: Correct

PR #149, opened 2018-08-17
feat(pie): only show pie slices when outerRadius >= innerRadius

T-Wizard:
**Checklist**

- [x] tests added
- [ ] documentation updated


packages/picasso.js/src/core/chart-components/pie/pie.js
```diff
@@ -167,7 +167,7 @@ const pieComponent = {
     let sum = 0;
     for (let i = 0, len = items.length; i < len; i++) {
       const val = arcValue(stngs, items[i]);
-      if (val > 0) {
+      if (val > 0 && items[i].outerRadius > items[i].innerRadius) {
```
miralemd: would prefer to have `>=` as comparison

PR #189, opened 2018-09-12
chore: publish source files

nilzona:
Publish source files so that `"module": "src/index.js"` works when transpiling picasso files as a npm dependency.


carlhannes:
Praise! @peol

PR #195, opened 2018-09-22
chore: move unit test closer to source file

nilzona:
This PR moves all unit tests into `/src` folder and can now be found in a `__test__` folder next the the files the tests are testings.


stoffeastrom:
image
Current master 👇 
image

stoffeastrom:
Also I think the folder should be named `__tests__` :)

nilzona:
@miralemd  that's exactly what I did if you look here .. https://github.com/qlik-oss/picasso.js/pull/195/files#diff-807e73f36c88f6bae3ad336c8d61cae9R2

But maybe the .eslintrc should be moved to the root instead of being referenced inside the `test-utils`package

miralemd:
aah, didn't see that (too many changes :P) - then yes, we should only have one eslint in the root

PR #248, opened 2018-11-02
feat: make it possible to exclude specific components from update

veinfors:
Makes it possible to send in an array called `excludeFromUpdate` in which you can specify the keys of the components you want to exclude from updating. This array is to be passed into `chartInstance.update`. This is one approach of optimizing performance (only re-render/update what's actually changed).

- [x] tests added
- [x] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -394,7 +394,7 @@ function chartFn(definition, context) {
    * @param {chart-definition} [chart] - Chart definition
    */
   instance.update = (newProps = {}) => {
-    const { partialData } = newProps;
+    const { partialData, excludeFromUpdate = [] } = newProps;
     if (newProps.data) {
       data = newProps.data;
     }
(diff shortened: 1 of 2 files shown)


packages/picasso.js/src/core/chart/index.js
```diff
@@ -426,6 +426,12 @@ function chartFn(definition, context) {
     // Let the "components" array determine order of components
     currentComponents = components.map((comp) => {
       const idx = findComponentIndexByKey(comp.key);
+
+      // Component should not be updated
+      if (excludeFromUpdate.length && excludeFromUpdate.indexOf(comp.key) > -1) {
```
cbt1: Is it intentional that you can exclude a component that is added the same update call? Based on your PR description, the logic doesn't seem to match.

cbt1: I would skip the length check. It's an array or it should throw. Similar to how we assume the `components` setting is an array.

veinfors: Done!

veinfors: It shouldn't be a problem since the component will be filtered out further down anyway.

PR #261, opened 2018-11-21
fix(dock-layout): remove components that are docked at hidden components

nguyenlam1:
**Checklist**

- [x] tests added
- [ ] documentation updated

Updates the visible and hidden components based on components that are docked to other components.
 For example, assume a component called myRect:
 {
  key: 'myRect',
   type: 'rect',
   dock: 'bottom'
 }
  and a component called myLine:
 {
  key: 'myLine',
  type: 'line',
 dock: '@myRect'
 }
 if the layout engine decides to hide myRect, then myLine should be hidden as well.

### DIFF
--- a/packages/picasso.js/src/core/dock-layout/dock-layout.js
+++ b/packages/picasso.js/src/core/dock-layout/dock-layout.js
@@ -124,15 +124,11 @@ function updateForComponentsDockedAtRemovedComponents(components, hiddenComponen
   const dockedAtRemovedComponents = [];
 
   const visibleComponents = components.filter((comp) => {
-    const dock = comp.config && comp.config.dock();
-
-    if (!/^@/.test(dock)) {
+    if (comp.referencedDocks.length === 0) {
       return true;
     }
 
-    const referencedDocks = dock.split(',').map(s => s.replace(/^\s*@/, ''));
-
-    const isAllHidden = referencedDocks.every(refDock => hiddenComponents.some((hiddenComp) => {
+    const isAllHidden = comp.referencedDocks.every(refDock => hiddenComponents.some((hiddenComp) => {
       const key = hiddenComp.ctx && hiddenComp.ctx.key;
       return refDock === key;
     }));
@@ -228,11 +224,13 @@ function positionComponents(components, logicalContainerRect, reducedRect, conta
   const referencedComponents = {};
   const referenceArray = components.slice();
   components.sort((a, b) => {
-    if (/^@/.test(b.config.dock())) {
-      return -1;
-    }
-    if (/^@/.test(a.config.dock())) {
-      return 1;
+    if (a.referencedDocks.length > 0 || b.referencedDocks.length > 0) {
+      if (/^@/.test(b.config.dock())) {
+        return -1;
+      }
+      if (/^@/.test(a.config.dock())) {
+        return 1;
+      }
     }
     const diff = a.config.displayOrder() - b.config.displayOrder();
     if (diff === 0) {
@@ -292,9 +290,8 @@ function positionComponents(components, logicalContainerRect, reducedRect, conta
         outerRect.width = rect.width = reducedRect.width;
         outerRect.height = rect.height = reducedRect.height;
     }
-    if (/^@/.test(d)) {
-      const spacesFollowedByAtRegex = /^\s*@/;
-      const refs = d.split(',').map(r => referencedComponents[r.replace(spacesFollowedByAtRegex, '')]).filter(r => !!r);
+    if (c.referencedDocks.length > 0) {
+      const refs = c.referencedDocks.map(r => referencedComponents[r]).filter(r => !!r);
       if (refs.length > 0) {
         outerRect = boundingBox(refs.map(r => r.outerRect));
         rect = boundingBox(refs.map(r => r.rect));
@@ -364,10 +361,19 @@ export default function dockLayout(initialSettings) {
     validateComponent(component);
     docker.removeComponent(component);
 
+    function getReferencedDocks(dock) {
+      if (!/^@/.test(dock)) {
+        return [];
+      }
+
+      return dock.split(',').map(s => s.replace(/^\s*@/, ''));
+    }
+
     components.push({
       instance: component,
       key,
-      config: component.dockConfig()
+      config: component.dockConfig(),
+      referencedDocks: getReferencedDocks(component.dockConfig().dock())
     });
   };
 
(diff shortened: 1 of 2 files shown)


packages/picasso.js/src/core/dock-layout/dock-layout.js
```diff
@@ -95,6 +95,48 @@ function reduceSingleLayoutRect(logicalContainerRect, reducedRect, edgeBleed, c,
...
+
+  if (hiddenComponents.length === 0) {
+    return { visibleComponents, dockedAtRemovedComponents: [] };
+  }
+
+  const dockedAtRemovedComponents = [];
+
+  hiddenComponents.forEach((hiddenComp) => {
+    components.forEach((comp, idx) => {
+      const dock = comp.instance && comp.instance.ctx && comp.instance.ctx.dock;
```
cbt1: Dock may be a function here. Should always read from the dockConfig.dock() to get the evaluated dock.

Use: `comp.dockConfig.dock()`

cbt1:
I think there are two issues to solve in order to make this work.
* Get the dock value of a component and find all possible target components, this should preferably be something that is only done once or/and ensure that it's resolved the same way everywhere the data is needed.
* Of the target components, which are still visible. If none, hide component, otherwise reduce the target components set to only the visible components.

packages/picasso.js/src/core/dock-layout/dock-layout.js
```diff
@@ -95,6 +95,67 @@ function reduceSingleLayoutRect(logicalContainerRect, reducedRect, edgeBleed, c,
...
+  if (hiddenComponents.length === 0) {
+    return { visibleComponents, dockedAtRemovedComponents: [] };
+  }
+
+  const dockedAtRemovedComponents = [];
+
+  components.forEach((comp, idx) => {
+    let nbrOfHidden = 0;
+    const dock = comp.config && comp.config.dock();
+    const referencedDocks = dock.split(',').map(s => s.replace(' ', ''));
```
cbt1: This would also threat a component with a dock string of `<some-key>,<some-other-key>` the same as the correct definition `@<some-key>,@<some-other-key>`

packages/picasso.js/src/core/dock-layout/dock-layout.js
```diff
@@ -95,6 +95,67 @@ function reduceSingleLayoutRect(logicalContainerRect, reducedRect, edgeBleed, c,
...
+      }
+    });
+
+    // Hide the component only if all the referenced components are hidden
+    if (nbrOfHidden === referencedDocks.length) {
+      visibleComponents.splice(idx, 1);
+      dockedAtRemovedComponents.push(comp.instance);
+    } else if (nbrOfHidden > 0) {
+      // Reassigning new dock target(s) if more that one of the referenced components are visible
+      comp.config.dock = () => newTargetComponents.join();
```
cbt1: `config.dock()` takes an optional parameter to re-assign the dock value. However, you cannot re-assign it at all, because you don't know if the user definition for the `dock` setting was a function or not.

packages/picasso.js/src/core/dock-layout/dock-layout.js
```diff
@@ -95,6 +95,67 @@ function reduceSingleLayoutRect(logicalContainerRect, reducedRect, edgeBleed, c,
...
+function updateForComponentsDockedAtRemovedComponents(components, hiddenComponents) {
+  const visibleComponents = components.slice();
+
+  if (hiddenComponents.length === 0) {
+    return { visibleComponents, dockedAtRemovedComponents: [] };
+  }
+
+  const dockedAtRemovedComponents = [];
+
+  components.forEach((comp, idx) => {
```
cbt1: Can you not just use a simple filter function? `components.filter(c => {
 if ('<if not a ref>') { return true; }
 const referencedDocks = '<get-the-dock-ref-ary>';
 return referencedDocks.every(refDock => aryWithHiddenKeys.some(key => key === 
 refDock));
});`

With that said, I just threw this together so it might not actually work..

packages/picasso.js/src/core/dock-layout/dock-layout.js
nguyenlam1: Will think of a fix to this.

packages/picasso.js/src/core/dock-layout/dock-layout.js
nguyenlam1: Removed

packages/picasso.js/src/core/dock-layout/dock-layout.js
nguyenlam1: Better with the refactored one? :)

nguyenlam1:
@cbt1 Ready for review again :)

packages/picasso.js/src/core/dock-layout/dock-layout.js
```diff
@@ -308,10 +361,19 @@ export default function dockLayout(initialSettings) {
     validateComponent(component);
     docker.removeComponent(component);
 
+    function getReferencedDocks(dock) {
```
cbt1: It's a pure function, there is no need to create this function every time a component is added.

packages/picasso.js/src/core/dock-layout/dock-layout.js
```diff
@@ -173,11 +224,13 @@ function positionComponents(components, logicalContainerRect, reducedRect, conta
   const referencedComponents = {};
   const referenceArray = components.slice();
   components.sort((a, b) => {
-    if (/^@/.test(b.config.dock())) {
-      return -1;
-    }
-    if (/^@/.test(a.config.dock())) {
-      return 1;
+    if (a.referencedDocks.length > 0 || b.referencedDocks.length > 0) {
```
cbt1: No need add an extra if check and than use the "old" reference check. Just do:
```js
if (b.referencedDocks.length > 0) {
  return -1;
}
if (a.referencedDocks.length > 0) {
  return 1;
}
```

nguyenlam1: Good point! Thanks!

packages/picasso.js/src/core/dock-layout/dock-layout.js
nguyenlam1: fixed!

packages/picasso.js/src/core/dock-layout/dock-layout.js
nguyenlam1: fixed!

PR #262, opened 2018-11-22
fix(events): turn off default event handlers

miralemd:
Removes the default event handlers when turning off interactions.

### DIFF
--- a/test/integration/interactions/interaction.int.js
+++ b/test/integration/interactions/interaction.int.js
@@ -1,9 +1,65 @@
+/* global picassochart, triggeredEvents */
+
+const iPad = {
+  name: 'iPad landscape',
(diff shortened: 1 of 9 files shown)


test/integration/interactions/interaction.int.js
```diff
@@ -1,9 +1,65 @@
...
+  const fixtureLasso = 'http://localhost:10001/interactions/interactions-lasso.fix.html';
   const fixtureRange = 'http://localhost:10001/interactions/interactions-range.fix.html';
 
+  describe('events', () => {
+    it('should be on by default', async () => {
+      await page.goto(fixture);
+      await page.waitForSelector('.container');
+      await page.click('.container');
+      const ev = await page.evaluate(() => triggeredEvents);
+      expect(ev.mousedown).to.equal(2);
```
cbt1: Slightly ode verification as you cannot be sure if an event was triggered twice or both the `on` and `native` was fired. But I do see the pain in setting up fixture to cover that case.

PR #280, opened 2019-01-08
feat(labels): allow text to overflow bars

cbt1:
This PR adds a new property on the labels components `bar` strategy. Enabling texts to be rendered even when they don't fully fit the bar.

There is a performance overhead introduced each label needs to be checked for possible intersections with other nodes and when all selected bars are sorted by direction.

The logic for determining the best possible placement doesn't consider overflow property for now.

Also should rename it to something other than overflow..

overflow-labels

**Usage:**
```js
{
        type: 'labels',
        settings: {
          sources: [
            {
              component: 'bars',
              selector: 'rect',
              strategy: {
                type: 'bar',
                settings: {
                  labels: [{
                    placements: [
                      { position: 'outside', fill: '#666', overflow: true }
                    ]
                  }]
                }
              },
            },
          ],
        },
      }
```

**Checklist**

- [x] tests added
- [x] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
+++ b/packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
@@ -2,6 +2,7 @@ import extend from 'extend';
 import {
   testRectRect
 } from '../../../math/narrow-phase-collision';
+import filterOverlapping from './bars-overlapping-filter';
 
 const LINE_HEIGHT = 1.5;
 const PADDING = 4;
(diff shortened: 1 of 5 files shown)


packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -17,6 +24,11 @@ function cbContext(node, chart) {
   };
 }
 
+function isValidText(text) {
+  const type = typeof text;
+  return (type === 'string' || type === 'number') && text !== '';
```
miralemd: I think only `type === 'string'` should be allowed

cbt1: I think a common case is where the user provide some data value that is a number. Ex: `({ data }) => data.someProp.value`.

miralemd: The return value should be coerced then (line 358)

PR #300, opened 2019-02-06
fix: setup emits on all components

nguyenlam1:
**Checklist**

- [x] tests added
- [ ] documentation updated

- instantiate all components with empty emit functions if they don't already have defined emitters.

### DIFF
--- a/packages/picasso.js/src/core/component/component-factory.js
+++ b/packages/picasso.js/src/core/component/component-factory.js
@@ -44,6 +44,8 @@ function prepareContext(ctx, definition, opts) {
     symbol
   } = opts;
 
+  ctx.emit = () => {};
+
   // TODO add setters and log warnings / errors to console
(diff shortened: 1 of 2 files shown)


packages/picasso.js/src/core/component/component-factory.js
```diff
@@ -145,6 +147,13 @@ function setUpEmitter(ctx, emitter, settings) {
   ctx.emit = (name, ...event) => emitter.emit(name, ...event);
 }
 
+function tearDownEmitter(ctx, emitter) {
+  (ctx.eventListeners || []).forEach(({ event, listener }) => {
+    emitter.removeListener(event, listener);
+    ctx.emit = () => {};
```
cbt1: No need to run this inside the forEach. There is only one ctx.

nguyenlam1: fixed!

PR #305, opened 2019-02-08
fix(svg-renderer): do not add undefined title attribute

miralemd:
Fixes `title='undefined'` in DOM output

Before:
```html
<text title="undefined" data-value="0" data-label="Americas">Americas</text>
```
After:
```html
<text data-value="0" data-label="Americas">Americas</text>
```

### DIFF
--- a/packages/picasso.js/src/core/scene-graph/display-objects/text.js
+++ b/packages/picasso.js/src/core/scene-graph/display-objects/text.js
@@ -57,7 +57,9 @@ export default class Text extends DisplayObject {
     this.attrs.dx = dx;
     this.attrs.dy = dy;
     this.attrs.text = text;
-    this.attrs.title = title;
+    if (typeof title !== 'undefined') {
+      this.attrs.title = String(title);
+    }
     if (typeof boundingRect === 'object') {
       this._textBoundsFn = () => boundingRect;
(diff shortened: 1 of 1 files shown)


packages/picasso.js/src/web/renderer/svg-renderer/svg-nodes.js
```diff
@@ -45,10 +45,12 @@ const maintainer = (element, item) => {
         val = item.attrs[attr];
       }
       element.setAttribute('dy', val + dy);
-    } else if (item.type === 'text' && attr === 'title' && item.attrs.title) {
-      const t = element.ownerDocument.createElementNS(svgNs, 'title');
-      t.textContent = item.attrs.title;
-      element.appendChild(t);
+    } else if (item.type === 'text' && attr === 'title') {
+      if (item.attrs.title) {
```
cbt1: do proper type check to support falsy values

miralemd: `item.attrs.title` should always be a string - I don't see the point in supporting an empty value here?

cbt1: I think you're putting a bit too much responsibility on the user.  The primary case would be to allow 0 as a number.

PR #310, opened 2019-02-13
fix(dock-layout): remove components that are docked at hidden components

nguyenlam1:
**Checklist**

- [x] tests added
- [ ] documentation updated

Updates the visible and hidden components based on components that are docked to other components.
 For example, assume a component called myRect:
```js
{
   key: 'myRect',
   type: 'rect',
   dock: 'bottom'
 }
```
  and a component called myLine:
```js
{
  key: 'myLine',
  type: 'line',
  dock: '@myRect'
 }
```
 if the layout engine decides to hide myRect, then myLine should be hidden as well.

### DIFF
--- a/packages/picasso.js/src/core/dock-layout/dock-layout.js
+++ b/packages/picasso.js/src/core/dock-layout/dock-layout.js
@@ -95,6 +95,54 @@ function reduceSingleLayoutRect(logicalContainerRect, reducedRect, edgeBleed, c,
   return true;
 }
 
+
+/**
+ * Updates the visible and hidden components based on components that are docked to other components.
+ * For example, assume a component called myRect:
+ * {
+ *  key: 'myRect',
+ *  type: 'rect',
+ *  dock: 'bottom'
+ * }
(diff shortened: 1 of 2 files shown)


packages/picasso.js/src/core/dock-layout/dock-layout.js
```diff
@@ -308,10 +359,13 @@ export default function dockLayout(initialSettings) {
...
     docker.removeComponent(component);
 
+    const dock = component.dockConfig().dock();
+
     components.push({
       instance: component,
       key,
-      config: component.dockConfig()
+      config: component.dockConfig(),
+      referencedDocks: /^@/.test(dock) ? dock.split(',').map(s => s.replace(/^\s*@/, '')) : []
```
miralemd: this won't work if the first dock isn't a reference, e.g.: `left,@points`

nguyenlam1: Good catch! Will fix! :)

PR #321, opened 2019-03-12
fix: update component config

veinfors:
This solves an issue when `preferredSize` has been updated in a component definition but the component keeps using the `preferredSize` specified in the initial definition.

This solution updated the config object in the `component-factory` when a component is updated. It might not be the optimal solution, or it may have unexpected effects... What do you guys think, @miralemd, @cbt1 ?


miralemd:
a test would be nice :)

PR #331, opened 2019-03-18
perf(q): switch for each loops to regular for loops 

nguyenlam1:
**Checklist**

- [ ] tests added
- [ ] documentation updated

- By switching for each loops to regular for loops inside extractor-s.js the following results was yielded when running performance benchmarks. This was tested on 500 field values.

for each loops
~ 2594 ops/s

for loops
~ 3763 ops/s

- In order to check out the metrics, please check out the branch located on this link and follow the instructions in README.md https://github.com/nguyenlam1/picasso.js/tree/perf-extractor-s/plugins/q/src/data

### DIFF
--- a/plugins/q/src/data/extractor-s.js
+++ b/plugins/q/src/data/extractor-s.js
@@ -97,74 +97,76 @@ function cellToValue({
 export default function extract(config, dataset, cache, util) {
   const cfgs = Array.isArray(config) ? config : [config];
   let dataItems = [];
-  cfgs.forEach((cfg) => {
-    if (typeof cfg.field !== 'undefined') {
+  for (let i = 0; i < cfgs.length; i++) {
+    if (typeof cfgs[i].field !== 'undefined') {
       const cube = dataset.raw();
       const sourceKey = dataset.key();
-      const f = typeof cfg.field === 'object' ? cfg.field : dataset.field(cfg.field);
-      const { props, main } = util.normalizeConfig(cfg, dataset);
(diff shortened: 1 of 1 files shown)


plugins/q/src/data/extractor-s.js
```diff
@@ -97,74 +97,76 @@ function cellToValue({
...
       const trackedItems = [];
       const items = [];
 
-      cube.qDataPages.forEach((page) => {
-        const fn = getFieldAccessor(f, page, { cache });
+      for (let j = 0; j < cube.qDataPages.length; j++) {
+        const fn = getFieldAccessor(f, cube.qDataPages[j], { cache });
         if (fn === -1) {
-          return;
+          continue;
```
nguyenlam1: The return did not do anything in the for each loop, so I replaced it with continue instead. If there is a better case here I that missed, please let me know.

plugins/q/src/data/extractor-s.js
```diff
@@ -97,74 +97,76 @@ function cellToValue({
...
-          const rowIdx = page.qArea.qTop + i;
-          const mainCell = extend({ qRow: rowIdx }, fn(row));
+        for (let k = 0; k < cube.qDataPages[j].qMatrix.length; k++) {
+          const rowIdx = cube.qDataPages[j].qArea.qTop + k;
+          const mainCell = extend({ qRow: rowIdx }, fn(cube.qDataPages[j].qMatrix[k]));
           const ret = datumExtract(main, mainCell, { key: sourceKey });
           const exclude = main.filter && !main.filter(mainCell);
           if (exclude) {
-            return;
+            continue;
```
nguyenlam1: Same as above

PR #335, opened 2019-03-22
perf(box): improve performance for boxes

quanho:
This PR is to:
1. only add markers (line, median, etc.) used,
2. optimize the loop of adding markers.


packages/picasso.js/src/core/chart-components/box/box-shapes.js
```diff
@@ -208,114 +209,174 @@ export function buildShapes({
...
+  let boxWidth;
+  let boxPadding;
+  let boxCenter;
+  let isLowerOutOfBounds;
+  let isHigherOutOfBounds;
+  let isOutOfBounds;
+  const numKeys = keys ? keys.length : 0;
+  const numNonOobKeys = nonOobKeys ? nonOobKeys.length : 0;
+
+  function addBox() {
```
cbt1: Why not go all the way and break out these "add" functions from buildShapes? As buildShapes is now a ~200 lines function with a lot of complex logic that should be tested separately.

packages/picasso.js/src/core/chart-components/box/box-shapes.js
```diff
@@ -208,114 +209,174 @@ export function buildShapes({
...
+        item: minorItem, value: 0, boxCenter, rendWidth, rendHeight, flipXY, symbol
+      }));
+    } else if (isHigherOutOfBounds) {
+      children.push(oob({
+        item: minorItem, value: 1, boxCenter, rendWidth, rendHeight, flipXY, symbol
+      }));
+    }
+  }
+
+  const addMarkerList = {
```
cbt1: Declare closer to usage.

cbt1:
Not a single new test?

quanho:
> Not a single new test?

Added unit tests.

PR #343, opened 2019-03-27
fix(bar-label): update checking if a label is inside a bar or not

quanho:


packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -29,6 +29,11 @@ function isValidText(text) {
   return (type === 'string' || type === 'number') && text !== '';
 }
 
+export function isTextInRect(rect, textMetrics, opts) {
+  return opts.rotate ? !(rect.width < opts.fontSize || rect.width < textMetrics.height || rect.height < textMetrics.width)
+    : !(rect.height < opts.fontSize || rect.width < textMetrics.width || rect.height < textMetrics.height);
```
cbt1: `rect.height < opts.fontSize` should be redundant in this case as `rect.height < textMetrics.height` as a more accurate expression of the same thing. Same for the rotate case I believe.

quanho: From @miralemd: The textMetrics by default may have only width but not height, so the fontSize is used for the height in that case. However, he think it is ok to remove the check against the fontSize. I will update the PR based on this.

cbt1: @quanho That is not the case. The textMetrics is used on multiple places in the code and check it's height property. When measure text, both width and height is always returned.

quanho: @cbt1: @miralemd  shown me the following  https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics

cbt1: @quanho Yes but that is not what the component is using. Not directly at least. It's using a "polyfilled" version that does provide height.

quanho: Good point!

PR #351, opened 2019-04-01
fix(bar-labels): fix the calculation of the bounds of a label when it is rotated -90 degree

quanho:


packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -193,21 +193,27 @@ function approxTextBounds(label, textMetrics, rotated, rect) {
...
   const width = rotated ? textMetrics.width : Math.min(textMetrics.width, rect.width);
   const height = rotated ? Math.min(textMetrics.height, rect.height) : textMetrics.height;
-  const PADDING_OFFSET = 0.1; // Needed to support a case when multiple bars are on the same location
+  const PADDING_OFFSET = 1e-6; // Needed to support a case when multiple bars are on the same location
 
-  const bounds = {
+  let bounds = {
     x: x - PADDING - PADDING_OFFSET,
-    y: y - height - PADDING - PADDING_OFFSET,
+    y: y - height * 0.75 - PADDING - PADDING_OFFSET,
```
miralemd: add a comment on how you came up with this magic number

quanho: It based on other magic numbers used in placeTextInRect.

miralemd: yeah, gonna need a bit more than "I just came up with it" - why does this number work and not others?

quanho: It will be a long comment:
text-before-edge

d3-base-line

Non-rotate:
The approximated text bound of a label is equivalent to the label when its baseline is 'text-before-edge'. However the label's baseline is 'alphabetical'. So we need to compensate the y-coordinate about 75% of the label's height.

miralemd: Nice!

PR #355, opened 2019-04-03
fix(bar-label): add padding and background color to bar labels

quanho:
This PR includes:
1. adding padding for bar labels which is used to calculate the label bound and check if the label bound is fit a bar or not,
2. adding background color to the label which is useful in case of showing stack total values when they are rendered inside bar segments.

image


quanho:
> the config for background needs to be documented and added to the API spec.

Added document for padding and background.

PR #354, opened 2019-04-03
refactor: prepare compose API

nilzona:
This PR is about preparing for the compose api described in #181  the dock layout settings in component settings has now been deprecated and they should be moved into a `layout` property.
A component definition that previously had:
```
{
   type: 'axis',
   dock: 'right',
   displayOrder: 3
}
```
Should instead be:
```
{
   type: 'axis',
   layout: {
      dock: 'right',
      displayOrder: 3
   }
}
```

Also dock layout engine has been moved and changed somewhat for preparing the compose api.

**Checklist**

- [x] tests added
- [x] commits conform to the [commit guidelines](./CONTRIBUTING.md#commit)
- [x] documentation updated

PR #362, opened 2019-04-05
docs: update dockLayout references

miralemd:
BREAKING CHANGE: rect is now a reserved keyword for components

## Breaking

A lot of examples used the `beforeRender` lifecycle method to set `this.rect` on the component:

```js
beforeRender(opts) {	
  this.rect = opts.size;	
},
```

💥 **`rect` is now a reserved keyword and automatically set on all components by the layout strategy**. Trying to assign a value to `rect` inside a component will throw an error.

> The actual breaking PR is https://github.com/qlik-oss/picasso.js/pull/354, but it did not contain any BREAKING CHANGE comment, so adding it in this commit instead.

## Deprecated

Properties related to the layout strategy have been moved one step down into a `layout` object, while `dockLayout` has been renamed to `strategy`:

**Before**
```js
picasso.chart({
  settings: {
    dockLayout: { // deprecated, use strategy instead
      logicalSize: { width: 400, height: 300 }
    },
    components: [{
      type: 'point',
      dock: 'left', // deprecated, use layout.dock instead 
      prioOrder: 1, // deprecated, use layout.prioOrder instead
      displayOrder: 2, // deprecated, use layout.displayOrder instead
      minimumLayoutMode: 'SMALL' // deprecated, use layout.minimumLayoutMode instead
    }]
  }
});
```

**After**
```js
picasso.chart({
  settings: {
    strategy: {
      logicalSize: { width: 400, height: 300 }
    },
    components: [{
      type: 'point',
      layout: {
        dock: 'left',
        prioOrder: 1,
        displayOrder: 2,
        minimumLayoutMode: 'SMALL'
      }
    }]
  }
});
```

### DIFF
--- a/docs/scriptappy.json
+++ b/docs/scriptappy.json
@@ -3,7 +3,7 @@
   "info": {
     "name": "picasso.js",
     "description": "A charting library streamlined for building visualizations for the Qlik Sense Analytics platform.",
-    "version": "0.22.0",
+    "version": "0.22.1",
     "license": "MIT"
   },
   "entries": {
@@ -780,7 +780,7 @@
               "type": "number"
             },
             "minimumLayoutMode": {
-              "description": "Refer to layout sizes defined by layoutModes in dockLayout",
+              "description": "Refer to layout sizes defined by layoutModes in `strategy`",
               "optional": true,
               "kind": "union",
               "items": [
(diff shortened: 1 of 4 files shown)


nilzona:
Maybe mention the change of dockLayout -> strategy property that is part of this PR?

PR #371, opened 2019-04-10
feat(brush): configure brush instance

cbt1:
Adds a new method to the brush instance api called `configure`. Through this method, it will be possible to configure various properties.

The first iteration will contain two properties that can control whether or not the min/max value of a range should be included when determining if a value is to be brushed.

**Example**
```js
brushInstance.configure({
   ranges: [
     { key: 'Sales' includeMax: false }, // Sales config
     { includeMax: true, includeMin: true }, // Default config
   ]
 })

brushInstance.addRange('Sales', { min: 20, max: 50 });
brushInstance.addRange('Margin', { min: 20, max: 50 });

brush.containsRangeValue('Sales', 50); // false
brush.containsRangeValue('Margin', 50); // true
```


packages/picasso.js/src/core/brush/brush.js
```diff
@@ -362,6 +392,48 @@ export default function brush({
...
+    if (Array.isArray(config.ranges) && config.ranges.length) {
+      rangeConfig = {
+        sources: {},
+        default: extend({}, DEFAULT_RANGE_CONFIG)
+      };
+
+      config.ranges.forEach((cfg) => {
+        const c = {};
+        Object.keys(DEFAULT_RANGE_CONFIG)
+          .filter(key => key !== 'key')
```
miralemd: should rename `key` to `attr` to be consistent with the next line and reduce confusion with `'key'`

T-Wizard:
It would be nice if q-brush.js would use this to set qMinInclEq and qMaxInclEq

cbt1:
> It would be nice if q-brush.js would use this to set qMinInclEq and qMaxInclEq

Interesting case! If it was to be automagacilly handled, then I think it would need to be tied the type of request method, which I think is out of scope for the brush.

miralemd:
`q-brush` would only need the configuration from the `brush`, so if we could expose `brush.config` it would be fairly straight forward to implement

cbt1:
@miralemd Still consider out of scope for this PR

PR #379, opened 2019-04-23
fix(legend-cat): fix setting initial legend scroll offset

quanho:
Problem:
When legend scroll offset is set before the data is set, the legend scroll offset will be reset to zero after the data is set (through getItemsToRender).
The problem comes from the fact that after the offset is set to zero, it is never reinitialized again by the checking offset === null in itemize function.

Solution:
Remove checking offset === null in the itemize function.

This will fix the problem of legend scroll offset is always at zero in the snapshot mode when a bar chart has 2 dimensions.

Before this fix:
image

After this fix:
image


quanho:
> test?

Added!

PR #381, opened 2019-04-24
fix(text-metrics): height measurement

cbt1:
Changes how the height of a text is measured to a method based on the font-size pixel value. This will make it less prone to issue with fonts that have narrow style for the character `M`.

See example below using `Arial` for the difference when computing the bounding rectangle of a text. Where the red rectangle is the bounds.

**Before**
Screenshot 2019-04-24 at 13 58 47

**After**
Screenshot 2019-04-24 at 13 58 33

### DIFF
--- a/packages/picasso.js/src/web/text-manipulation/font-size-to-height.js
+++ b/packages/picasso.js/src/web/text-manipulation/font-size-to-height.js
@@ -2,7 +2,7 @@ const BASE = 24;
 const PAD = 4;
 const BUMP = 1e-12;
 const DEFAULT_HEIGHT = 16;
-const TEXT_REGEX = /^\s*([0-9]+|[0-9]+\.{1}[0-9]+)[px]{2}\s*$/i;
+const TEXT_REGEX = /^\s*\d+(\.\d+)?px\s*$/i;
 
 function isValidFontSize(val) {
   const type = typeof val;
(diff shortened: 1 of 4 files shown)


packages/picasso.js/src/web/text-manipulation/font-size-to-height.js
```diff
@@ -0,0 +1,25 @@
+const BASE = 24;
+const PAD = 4;
+const BUMP = 1e-12;
+const DEFAULT_HEIGHT = 16;
+const TEXT_REGEX = /^\s*([0-9]+|[0-9]+\.{1}[0-9]+)[px]{2}\s*$/i;
```
miralemd: simplify the regex to `/^\s*\d(\.\d+)?px\s*/`

packages/picasso.js/src/web/text-manipulation/font-size-to-height.js
```diff
@@ -0,0 +1,25 @@
...
+
+function isValidFontSize(val) {
+  const type = typeof val;
```
miralemd: could write all of this as `return typeof value === 'string' ? TEST_REGEX.test(val) : false`

packages/picasso.js/src/web/text-manipulation/font-size-to-height.js
cbt1: Doesn't work for multiple number, ex: 90px. But will fix n update.

cbt1: Also regarding case, I've tried to follow the browser implementation for SVG text element. Where in FF, chrome and safari, PX and px is allowed.

PR #383, opened 2019-04-24
fix(legend-cat): add enabled/disabled state to navigation buttons

quanho:
This PR is to allow disabling navigation buttons of a categorical color legend.


quanho:
> add `disabled` to the spec: https://github.com/qlik-oss/picasso.js/blob/master/packages/picasso.js/src/core/chart-components/legend-cat/legend-resolver.js#L94

Added!

PR #390, opened 2019-04-29
fix: brush based on the amount of components

nguyenlam1:
**Checklist**

- [x] tests added
- [ ] documentation updated

- brush from shapes based on the amount of components sent in. Basically allows multiple components sharing the same key, but different contexts to be brushed on.

### DIFF
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -747,12 +747,9 @@ function chartFn(definition, context) {
   instance.brushFromShapes = (shapes, config = { components: [] }) => {
     for (let i = 0; i < config.components.length; i++) {
       const iKey = config.components[i].key;
-      visibleComponents.forEach((c) => {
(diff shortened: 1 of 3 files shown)


packages/picasso.js/src/core/chart/index.js
```diff
@@ -745,20 +745,16 @@ function chartFn(definition, context) {
...
-      if (configIndex !== -1) {
-        const compShapes = [];
-        for (let i = 0, num = shapes.length; i < num; i++) {
-          const shape = shapes[i];
-          if (shape.key === c.key) {
-            compShapes.push(shape);
-          }
+    for (let i = 0; i < config.components.length; i++) {
+      const iKey = config.components[i].key;
+      visibleComponents.forEach((c) => {
```
cbt1: I would simplify this to `visibleComponets.filter(c => iKey === c.key).forEach(...)`

nguyenlam1: fixed!

PR #444, opened 2019-09-26
fix(bar-label): take overflow into finding best placement for bar label

quanho:
Overflow property is used to ignore checking if a label is fit into a rectangle.
So we should take this into account of finding best placement for labels.

### DIFF
--- a/packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
+++ b/packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
@@ -32,9 +32,29 @@ function toBackground(label) {
   };
 }
 
-export function isTextInRect(rect, textMetrics, opts) {
-  return opts.rotate ? !(rect.width < textMetrics.height || rect.height < textMetrics.width)
-    : !(rect.width < textMetrics.width || rect.height < textMetrics.height);
+function isFitWidth(rect, label, rotate, ignoreCheck) {
+  return ignoreCheck || (rotate ? rect.width >= label.height : rect.width >= label.width);
+}
+
+function isFitHeight(rect, label, rotate, ignoreCheck) {
+  return ignoreCheck || (rotate ? rect.height >= label.width : rect.height >= label.height);
+}
+
+function isTextFitRect(orientation, rect, label, fitsHorizontally, overflow) {
+  let fitWidth;
+  let fitHeight;
+  if (orientation === 'v') {
+    fitWidth = fitsHorizontally || isFitWidth(rect, label, !fitsHorizontally, overflow);
+    fitHeight = isFitHeight(rect, label, !fitsHorizontally);
+  } else {
+    fitWidth = isFitWidth(rect, label);
+    fitHeight = isFitHeight(rect, label, false, overflow);
+  }
+  return fitWidth && fitHeight;
+}
+
+export function isTextInRect(rect, label, opts) {
+  return isFitWidth(rect, label, opts.rotate) && isFitHeight(rect, label, opts.rotate);
 }
 
 export function placeSegmentInSegment(majorSegmentPosition, majorSegmentSize, minorSegmentSize, align) {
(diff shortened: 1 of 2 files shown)


packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -167,10 +167,10 @@ export function findBestPlacement({
     largest = !p || testBounds.height > largest.height ? testBounds : largest;
 
     if (orientation === 'v' && ((fitsHorizontally && testBounds.height >= measured.height)
-      || (!fitsHorizontally && testBounds.height >= measured.width && testBounds.width >= measured.height))) {
+      || (!fitsHorizontally && testBounds.height >= measured.width && (!!placement.overflow || testBounds.width >= measured.height)))) {
```
cbt1: Let's not write nested if statements like this. Break each condition down into its own statement similar to:

```js
if (orientation === 'v') {
      if (fitsHorizontally) {
        if (testBounds.height >= measured.height) {
          bounds = testBounds;
          break;
        }
      } else if (testBounds.height >= measured.width && testBounds.width >= measured.height) {
        bounds = testBounds;
        break;
      }
    }
```

Or even better. Find a way to reduce the complexity of this function!

quanho: Refactored!

packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -32,9 +32,29 @@ function toBackground(label) {
...
-    : !(rect.width < textMetrics.width || rect.height < textMetrics.height);
+function isFitWidth(rect, label, rotate, ignoreCheck) {
+  return ignoreCheck || (rotate ? rect.width >= label.height : rect.width >= label.width);
+}
+
+function isFitHeight(rect, label, rotate, ignoreCheck) {
+  return ignoreCheck || (rotate ? rect.height >= label.width : rect.height >= label.height);
+}
+
+function isTextFitRect(orientation, rect, label, fitsHorizontally, overflow) {
```
cbt1: So what does isTextFitRect mean?

cbt1: And how does it differ from isTextInRect?

quanho: isTextFitRect  takes different types of parameter than isTextInRect. In core, they will use the same things (isFitWidth and isFitHeight).

packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -32,9 +32,29 @@ function toBackground(label) {
...
+  let fitWidth;
+  let fitHeight;
+  if (orientation === 'v') {
+    fitWidth = fitsHorizontally || isFitWidth(rect, label, !fitsHorizontally, overflow);
```
cbt1: I don't think it adds any value to pass in the `overflow` variable to "isFitXXXX". It would be easier to read if it parts of this statement instead. Such that `fitWidth = fitsHorizontally || overflow || isFitWidth(rect, label, !fitsHorizontally);`

It would also skip one computation when the overflow parameter is not passed in to the function.

packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
cbt1: I do realize that naming is very hard, but `isTextFitRect, isFitWidth and isFitHeight` doesn't really mean anything. So I suggest you try and be a bit more descriptive and name them after their purpose in plain english.

packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
quanho: Fixed!

packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
quanho: Updated!

PR #447, opened 2019-10-08
fix(labels): consider orientation when determining largest rect

quanho:


packages/picasso.js/src/core/chart-components/labels/strategies/__tests__/bars.spec.js
```diff
@@ -253,6 +255,22 @@ describe('labeling - bars', () => {
       expect(p.bounds).to.equal(rects.biggest);
     });
 
+    it('should find largest rect as fallback, horizontal, and when all rects have the same height', () => {
+      rects.inside.height = rects.outside.height = rects.biggest.height = rects.meh.height = 100;
```
cbt1: Don't use this kind of syntax. Declare a const with desired value and assign it to each property on a separate row.

packages/picasso.js/src/core/chart-components/labels/strategies/__tests__/bars.spec.js
```diff
@@ -267,6 +285,22 @@ describe('labeling - bars', () => {
       expect(p.bounds).to.equal(rects.biggest);
     });
 
+    it('should find largest rect as fallback, vertical, and when all rects have the same width', () => {
+      rects.inside.width = rects.outside.width = rects.biggest.width = rects.meh.width = 100;
```
cbt1: Don't use this kind of syntax. Declare a const with desired value and assign it to each property on a separate row.

packages/picasso.js/src/core/chart-components/labels/strategies/__tests__/bars.spec.js
quanho: Fixed!

packages/picasso.js/src/core/chart-components/labels/strategies/__tests__/bars.spec.js
quanho: Fixed!

PR #453, opened 2019-11-12
feat(axis): allow setting paddingStart and paddingEnd as either a func or a number

johanlahti:
**Checklist**

- [x] tests added
- [x] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/chart-components/axis/axis.js
+++ b/packages/picasso.js/src/core/chart-components/axis/axis.js
@@ -22,10 +22,13 @@ function alignTransform({ align, inner }) {
 function resolveAlign(align, dock) {
   const horizontal = ['top', 'bottom'];
   const vertical = ['left', 'right'];
-  if (horizontal.indexOf(align) !== -1 && vertical.indexOf(dock) === -1) {
-    return align;
-  }
-  if (vertical.indexOf(align) !== -1 && horizontal.indexOf(dock) === -1) {
+
+  const docksHorizontally = horizontal.includes(dock);
+  const docksVertically = vertical.includes(dock);
+  const alignsHorizontally = horizontal.includes(align);
+  const alignsVertically = vertical.includes(align);
+
+  if ((alignsHorizontally && !docksVertically) || (alignsVertically && !docksHorizontally)) {
     return align;
   }
   return dock; // Invalid align, return current dock as default
@@ -38,7 +41,8 @@ function resolveAlign(align, dock) {
 function resolveLocalSettings({
   state,
   style,
-  settings
+  settings,
+  chart
 }) {
   const defaultStgns = extend(true, {}, state.isDiscrete ? DEFAULT_DISCRETE_SETTINGS : DEFAULT_CONTINUOUS_SETTINGS, style);
   const localStgns = extend(true, {}, defaultStgns, settings.settings);
@@ -47,6 +51,9 @@ function resolveLocalSettings({
   localStgns.dock = dock;
   localStgns.align = resolveAlign(settings.settings.align, dock);
   localStgns.labels.tiltAngle = Math.max(-90, Math.min(localStgns.labels.tiltAngle, 90));
+
+  const { paddingStart } = localStgns;
+  localStgns.paddingStart = typeof paddingStart === 'function' ? paddingStart.call(null, chart, state) : paddingStart;
 
   return localStgns;
 }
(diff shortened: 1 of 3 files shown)


qlikossbuild:
Hi, and thanks for contributing!

Note: If you are part of Qlik R&D, you can ignore this process (but let us know in a comment below!)

Before we can handle your pull request we need you to sign our contributor license agreement.
This is a one-time thing and enables you to contribute without hassle to any of the qlik-oss repositories in the future.

And it only takes a few minutes! Here are some useful links to get you started:

* [Sign the CLA](https://github.com/qlik-oss/open-source/blob/master/sign-cla.md#sign-the-cla)
* [Rationale behind the CLA](https://github.com/qlik-oss/open-source/blob/master/cla-rationale.md)
* [Qlik R&D Open Source repository](https://github.com/qlik-oss/open-source#open-source-at-qlik-rd)
* [Official Slack team](http://branch.qlik.com/slack)

Having troubles? Feel free to reach out to us either here or on Slack!

packages/picasso.js/src/core/chart-components/axis/axis.js
```diff
@@ -22,10 +22,13 @@ function alignTransform({ align, inner }) {
 function resolveAlign(align, dock) {
   const horizontal = ['top', 'bottom'];
   const vertical = ['left', 'right'];
-  if (horizontal.indexOf(align) !== -1 && vertical.indexOf(dock) === -1) {
-    return align;
-  }
-  if (vertical.indexOf(align) !== -1 && horizontal.indexOf(dock) === -1) {
+
+  const docksHorizontally = horizontal.includes(dock);
```
miralemd: Tests on IE11 are failing most likely because of this: we don't have a polyfill for `includes`

packages/picasso.js/src/core/chart-components/axis/axis.js
```diff
@@ -48,6 +52,9 @@ function resolveLocalSettings({
   localStgns.align = resolveAlign(settings.settings.align, dock);
   localStgns.labels.tiltAngle = Math.max(-90, Math.min(localStgns.labels.tiltAngle, 90));
 
+  const { paddingStart } = localStgns;
+  localStgns.paddingStart = typeof paddingStart === 'function' ? paddingStart.call(null, chart, state) : paddingStart;
```
miralemd: exposing `state` which is an internal variable containing tons of stuff is a big nono - don't expose more than you need for the job. Typically the first param is an object containing the minimum amount of properties needed for the intended purpose, that way we reduce the risk of breaking changes if we need more things.

cbt1: The parameters passed to the callback function should ideally also follow a similar pattern across all components. Typically it looks like this:
```js
paddingStart: ({
  resources: { scale, formatter }, // scale and formatter is the same function exposed by the chart instance
  /*  additional properties that may be included here such as the scale used by the component and/or the dataset and/or a datum */
}) => {}
```

In your case I believe it could be enough with the above example.

packages/picasso.js/src/core/chart-components/axis/axis.js
```diff
@@ -22,10 +22,13 @@ function alignTransform({ align, inner }) {
...
-    return align;
-  }
-  if (vertical.indexOf(align) !== -1 && horizontal.indexOf(dock) === -1) {
+
+  const docksHorizontally = horizontal.includes(dock);
+  const docksVertically = vertical.includes(dock);
+  const alignsHorizontally = horizontal.includes(align);
+  const alignsVertically = vertical.includes(align);
+
+  if ((alignsHorizontally && !docksVertically) || (alignsVertically && !docksHorizontally)) {
```
cbt1: I thinking this could be simplified to something like this:
```js
const validAlign = vertical.some(v === dock) && vertical.some(v === align);

return validAlign ? align : dock;
```

packages/picasso.js/src/core/chart-components/axis/axis.js
johanlahti: Ok. I only have to reach `state.innerRect` and the major `scale`. Feels wrong to send that as a specific parameter (`innerRect`)?

packages/picasso.js/src/core/chart-components/axis/axis.js
johanlahti: Ok, I should probably revert that whole thing actually.

packages/picasso.js/src/core/chart-components/axis/axis.js
cbt1: At the time of resolving the settings. The size of the component hasn't be fully resolved yet. So the "rect" that you could pass down as a parameter wouldn't be accurate in all cases.

You best bet here is to use closure to access a guess-estimated size.

packages/picasso.js/src/core/chart-components/axis/axis.js
johanlahti: I think I will leave that refactoring part for this PR.

PR #456, opened 2019-11-14
feat(point): add control to show/hide a point in a point component

quanho:


packages/picasso.js/src/core/chart-components/point/point.js
```diff
@@ -101,7 +101,7 @@ function getPointSizeLimits(x, y, width, height, limits) {
 function createDisplayPoints(dataPoints, {
   width, height
 }, pointSize, shapeFn) {
-  return dataPoints.filter((p) => !isNaN(p.x + p.y)).map((p) => {
+  return dataPoints.filter((p) => p.show !== false && !isNaN(p.x + p.y)).map((p) => {
```
miralemd: test? docs?

PR #468, opened 2019-12-11
refactor: adjust tabindex of legend nav

veinfors:
Setting tabindex to -1 for legend navigation buttons to improve accessibility/keyboard navigation experienvce.

### DIFF
--- a/packages/picasso.js/src/core/chart-components/legend-cat/navigation-renderer.js
+++ b/packages/picasso.js/src/core/chart-components/legend-cat/navigation-renderer.js
@@ -110,7 +110,8 @@ function render(renderer, {
     direction: order[0],
     attrs: {
       'data-action': 'next',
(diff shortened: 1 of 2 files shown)


packages/picasso.js/src/core/chart-components/legend-cat/navigation-renderer.js
```diff
@@ -110,7 +110,8 @@ function render(renderer, {
     direction: order[0],
     attrs: {
       'data-action': 'next',
-      'data-component-key': legend.settings.key
+      'data-component-key': legend.settings.key,
+      tabindex: -1
```
miralemd: I would prefer if this was configurable to keep it accessible

veinfors: Fixed!

PR #470, opened 2019-12-18
feat(geometry): add multipolygon

quanho:
This PR is to:
1. add multi-polygons which may have holes
2. to solve the tooltip and selection problem for donut chart when there is only one value
3. fix checking intersection between two polygons


packages/picasso.js/src/core/geometry/geopolygon.js
```diff
@@ -0,0 +1,225 @@
...
+ * A geo-polygon is a polygon which is similar to a polygon in GeoJson. A typical geopolygon is an array of polygons where the first polygon is an outer polygon and the rest are inner polygons
+ * @private
+ */
+
+/**
+ * Construct a new GeoPolygon instance
+ * @private
+ */
+class GeoPolygon {
+  constructor({ vertices = [[]] } = {}) {
```
cbt1: I don't think naming the variable `vertices` make any sense. It's an array of "polygons" or an array of "polygon definitions". It might be hard to find a good name for it. I'd suggest the one of the following: ´data´ or ´polygons´

packages/picasso.js/src/core/geometry/geopolygon.js
```diff
@@ -0,0 +1,225 @@
...
+      return testPolygonGeoPolygon(polygon, this);
+    }
+    return testGeoPolygonPolygon(this, polygon);
+  }
+
+  /**
+   * Get the points
+   * @returns {point[]}
+   */
+  points() {
```
cbt1: `points` API should return an array of points. The current use case for it is when debugging collision logic. I suggest you just return an empty array to avoid exceptions.

cbt1:
There are a couple of things missing in this PR to make it work in picasso chart:
- All geometry classes needs to have a method called `intersectsGeoPolygon`
- It should be possible query the `shapesAt` API with a geo-polygon (ex: `chartInstance.shapesAt(MyGeoPolygonObject)`. Add support for this in collision-resolver.js.

packages/picasso.js/src/core/geometry/geopolygon.js
quanho: This naming style respects GeoPolygon style..

packages/picasso.js/src/core/geometry/geopolygon.js
quanho: This is the same as polygon.

quanho:
> There are a couple of things missing in this PR to make it work in picasso chart:
> 
> * All geometry classes needs to have a method called `intersectsGeoPolygon`
> * It should be possible query the `shapesAt` API with a geo-polygon (ex: `chartInstance.shapesAt(MyGeoPolygonObject)`. Add support for this in collision-resolver.js.
> * display-object.js should have a method called `intersectsGeoPolygon`

Added!

packages/picasso.js/src/core/geometry/geopolygon.js
cbt1: It's still invalid.

PR #482, opened 2020-02-28
feat(brush): suppress paint on brush.end() and brush.start()

nilzona:
Add possibility to suppress picasso component render when doing brush.start and brush.end

- [x] tests added
- [x] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/brush/__tests__/brush.spec.js
+++ b/packages/picasso.js/src/core/brush/__tests__/brush.spec.js
@@ -39,11 +39,11 @@ describe('brush', () => {
   });
 
   describe('events', () => {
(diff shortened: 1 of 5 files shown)


packages/picasso.js/src/core/brush/__tests__/brush.spec.js
```diff
@@ -39,11 +39,11 @@ describe('brush', () => {
   });
 
   describe('events', () => {
-    it('should emit a start event when started', () => {
+    it('should emit a start event with paremeters when started', () => {
```
miralemd:
```suggestion
    it('should emit a start event with parameters when started', () => {
```

PR #485, opened 2020-03-16
fix(bar-label): allow fill function accessing label bound to support checking overlapping

quanho:
There is a need to decide the color of a label depending on if the label is overlapping with other objects. To be able to do that, in the fill color function, we need to know the label bound to support checking overlapping between the label and other objects.

### DIFF
--- a/packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
+++ b/packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
@@ -307,7 +307,7 @@ export function placeInBars(
 
         if (label) {
           const textBounds = approxTextBounds(label, measured, isRotated, bounds, placement.padding);
-          fill = typeof placement.fill === 'function' ? placement.fill(arg, i, textBounds) : placement.fill;
+          fill = typeof placement.fill === 'function' ? placement.fill({ ...arg, textBounds }, i) : placement.fill;
           label.fill = fill;
           if (typeof linkData !== 'undefined') {
             label.data = linkData;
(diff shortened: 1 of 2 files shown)


packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -308,13 +306,16 @@ export function placeInBars(
         });
 
         if (label) {
+          const textBounds = approxTextBounds(label, measured, isRotated, bounds, placement.padding);
+          fill = typeof placement.fill === 'function' ? placement.fill(arg, i, textBounds) : placement.fill;
```
cbt1: Don't like adding `textBounds` as a 3rd argument. A better solution IMO would be to extend the first argument with the `textBounds`

packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -308,13 +306,16 @@ export function placeInBars(
...
+          label.fill = fill;
           if (typeof linkData !== 'undefined') {
             label.data = linkData;
           }
           if (typeof placement.background === 'object') {
             label.backgroundColor =
               typeof placement.background.fill === 'function'
-                ? placement.background.fill(arg, i)
+                ? placement.background.fill(arg, i, textBounds)
```
cbt1: Same comment as above regarding 3rd argument

quanho: Fixed!

packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
quanho: Fixed!

PR #510, opened 2020-08-04
feat: support show/hide ref-line value

quanho:


packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -101,10 +101,12 @@ export function createLineWithLabel({ chart, blueprint, renderer, p, settings, i
       formatter = scaleData && scaleData[0] ? scaleData[0].formatter() : null;
     }
 
-    if (formatter) {
-      valueString = ` (${formatter(p.value)})`;
-    } else if (p.scale) {
-      valueString = ` (${p.value})`;
+    if (p.showValue !== false) {
```
miralemd, line 104: I'd rather we expand the use of `value` here to allow it to be set to `null`, and then only show it when `value != null`.

quanho: But we need value to calculate position. So we cannot set value to null unfortunately.

PR #516, opened 2020-08-18
feat(point): add direction setting

jjxie:
**Checklist**

- [ ] tests added
- [ ] documentation updated

Add direction setting to point. For example for horizontal combo chart, triangle and line markers can be rotated.


cbt1:
I don't think this solution is generic enough. It solves a specific issue, in your case rotating some shapes at a specific angle. But won't for really help in a broader sense. There are a number of properties that could be applied to the shapes available in picasso, `direction` is only applicable for a handful of those and the value that should be set is also different `vertical|horizontal` or `up|down|left|right`.

What we need is a solution that can take any shape property and pass it along. Such that all existing "shapes" (read symbols) in picasso can be fully configured and it would allow custom symbols to be created with custom properties.

cbt1:
See comment above

PR #542, opened 2020-11-02
fix: avoid problem during tree data extraction

veinfors:
Additional checks to avoid problem during data extraction for tree data nodes that are missing qAttrDims or qAttrExps.

### DIFF
--- a/plugins/q/src/data/__tests__/extractor-t.spec.js
+++ b/plugins/q/src/data/__tests__/extractor-t.spec.js
@@ -929,13 +929,13 @@ describe('q-data-extractor-t', () => {
         deps
       );
 
+      // Restore
+      treePages[0].qNodes[0].qAttrDims = tempQAttrDims;
+
       expect(m).to.eql([
         { value: NaN, label: '', source: { key: 'cube', field: 'qDimensionInfo/0/qAttrDimInfo/1' } },
         { value: 7, label: 'Bake', source: { key: 'cube', field: 'qDimensionInfo/0/qAttrDimInfo/1' } },
       ]);
-
-      // Restore
-      treePages[0].qNodes[0].qAttrDims = tempQAttrDims;
     });
 
     it('should be fine with nodes missing qAttrExps', () => {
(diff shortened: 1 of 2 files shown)


T-Wizard:
when is this needed?

veinfors:
> when is this needed?

When fetching data using `GetHyperCubeTreeData` and there is a data set which doesn't contain all possible dimension values in all "parent" nodes it's possible to send in a parameter called `AllValues` to engine. This tells engine to "pad"/"expand" with all possible values from the other dimension. But when doing this, engine is (of course) not adding values or attribute dimensions or attribute expressions on those nodes, since they don't exist. This PR makes it possible to use that kind of data set with picasso without it crashing.

T-Wizard (approved):

plugins/q/src/data/__tests__/extractor-t.spec.js
```diff
@@ -914,6 +914,56 @@ describe('q-data-extractor-t', () => {
...
+      expect(m).to.eql([
+        { value: NaN, label: '', source: { key: 'cube', field: 'qDimensionInfo/1/qAttrExprInfo/0' } },
+        { value: 2, label: 'two', source: { key: 'cube', field: 'qDimensionInfo/1/qAttrExprInfo/0' } },
+        { value: 5, label: 'five', source: { key: 'cube', field: 'qDimensionInfo/1/qAttrExprInfo/0' } },
+        { value: 7, label: 'seven', source: { key: 'cube', field: 'qDimensionInfo/1/qAttrExprInfo/0' } },
+        { value: 9, label: 'nine', source: { key: 'cube', field: 'qDimensionInfo/1/qAttrExprInfo/0' } },
+      ]);
+
+      // Restore
+      treePages[0].qNodes[0].qNodes[0].qAttrExps = tempQAttrExps;
```
cbt1, line 965: This wont be called if the `expect` ☝️ fails

cbt1 (changes requested):
Fix restore

plugins/q/src/data/__tests__/extractor-t.spec.js
veinfors: Good point! Fixed!

cbt1 (approved):

plugins/q/src/data/__tests__/extractor-t.spec.js
T-Wizard: grouping the test with a describe and moving this to an after/afterEach function will also run it if there is a error during the extract call.

PR #556, opened 2021-02-08
feat: apply transform to rendered components

veinfors:
This makes it possible to trigger a "light" rendering of specific components, where only a transform is applied. The reason is to provide smoother rendering performance for charts containing components with a lot of nodes. This approach might be applicable while interacting with a chart where you want to redraw the chart continuously and maintain a responsive feeling.

Here is how it is used:
```
chart.update({
  transforms[{
    key: 'component-key',
    transform: { a: 1, b: 0, c: 0, d: 1, e: 100, f: 200 },
  }]
});
```

But a component rendering using the canvas-renderer needs to provide rendererSettings in the definition (bufferSize is optional) like:
```
rendererSettings: {
  useBuffer: true,
  canvasBufferSize: {
    width: 1000,
    height: 800,
  },
},
```

Here is an example of how it can affect performance (a simple scatterplot with 2000 nodes) where chart rendering time went from 80 ms to 5 ms (the bottom one uses transfoms):
transfoms

**Checklist**

- [ ] tests added
- [ ] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -513,7 +513,7 @@ function chartFn(definition, context) {
    * @param {chart-definition} [chart] - Chart definition
    */
   instance.update = (newProps = {}) => {
-    const { partialData, excludeFromUpdate = [] } = newProps;
+    const { partialData, excludeFromUpdate = [], transforms = [] } = newProps;
     let visibleOrder;
     if (newProps.data) {
       data = newProps.data;
@@ -549,6 +549,11 @@ function chartFn(definition, context) {
           return currentComponents[idx];
         }
 
+        if (transforms.some((t) => t.key === comp.key)) {
+          currentComponents[idx].transform = transforms.filter((t) => t.key === comp.key)[0].transform;
+          return currentComponents[idx];
+        }
+
         if (idx === -1) {
           // Component is added
           return createComponent(comp, element);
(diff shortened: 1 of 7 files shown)


quanho:
> This makes it possible to trigger a "light" rendering of specific components, where only a transform is applied. The reason is to provide smoother rendering performance for charts containing components with a lot of nodes. This approach might be applicable while interacting with a chart where you want to redraw the chart continuously and maintain a responsive feeling.
> 
> Here is how it is used:
> 
> ```
> chart.update({
>   transforms[{
>     key: 'component-key',
>     transform: { a: 1, b: 0, c: 0, d: 1, e: 100, f: 200 },
>   }]
> });
> ```
> 
> But a component rendering using the canvas-renderer needs to provide rendererSettings in the definition (bufferSize is optional) like:
> 
> ```
> rendererSettings: {
>   useBuffer: true,
>   canvasBufferSize: {
>     width: 1000,
>     height: 800,
>   },
> },
> ```
> 
> Here is an example of how it can affect performance (a simple scatterplot with 2000 nodes) where chart rendering time went from 80 ms to 5 ms (the bottom one uses transfoms):
> transfoms
> 
> **Checklist**
> 
> * [ ]  tests added
> * [ ]  commits conform to the [commit guidelines](./CONTRIBUTING.md#commit)
> * [ ]  documentation updated

Very impressive performance improvement with this technique!

PR #566, opened 2021-03-15
fix(ref-line): wrong line position when scale min/max are equal

quanho:


packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -196,7 +196,8 @@ const refLineComponent = {
 
       if (line.scale) {
         let scale = this.chart.scale(line.scale);
-        return extend(line, { scale, position: scale(line.value) });
+        const position = scale.min() === scale.max() && line.value !== scale.min() ? NaN : scale(line.value);
```
cbt1, line 199: As evident by the unit-test you have added, this will output nodes that have `NaN` coordinates. Which only works using SVG renderer because of a fix I made few weeks ago. It's a pattern we should avoid using even tho it happens  to work.

So add a condition to the filter functions on line 222 and 230 where the nodes are removed if the position is `NaN`.

packages/picasso.js/src/core/chart-components/ref-line/__tests__/refline.spec.js
```diff
@@ -230,6 +234,99 @@ describe('reference lines', () => {
     ]);
   });
 
+  it('should render basic line with RTL label on X with scale when scale.min = scale.max', () => {
```
cbt1, line 237: Remove the "RTL" part of this test. Including the RTL in the config. It adds no value to the test of this logic.

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -20,6 +20,27 @@ function filterUndefinedValue(line) {
   return typeof value !== 'undefined';
 }
 
+function isInvert(scale) {
+  const range = scale.range();
+  return range?.length === 2 && range[0] === 1 && range[1] === 0;
```
cbt1, line 25: There is no guarantee that the range is 0 and 1. So I think you have to check if the first value is larger then the last value.

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -20,6 +20,27 @@ function filterUndefinedValue(line) {
...
+}
+
+function getPosition(scale, value) {
+  const min = scale.min();
+  const max = scale.max();
+  if (min === max) {
+    const invert = isInvert(scale);
+    console.log(invert);
```
cbt1, line 33: Remove log call

packages/picasso.js/src/core/chart-components/ref-line/refline.js
quanho: ok

cbt1 (approved):

PR #568, opened 2021-03-17
chore: add build:dev commands

quanho:


packages/picasso.js/package.json
```diff
@@ -27,6 +27,7 @@
   "module": "dist/picasso.esm.js",
   "scripts": {
     "build": "cross-env NODE_ENV=production rollup --config ../../rollup.config.js",
+    "build:dev": "cross-env rollup --config ../../rollup.config.js",
```
T-Wizard, line 30:
```suggestion
    "build": "rollup --config ../../rollup.config.js",
```
I would suggest removing NODE_ENV=production here. As i think it is enough to have in the top level package.json
(same for the package.json in hammer and q plugins)

plugins/hammer/package.json
```diff
@@ -15,6 +15,7 @@
   "module": "dist/picasso-hammer.esm.js",
   "scripts": {
     "build": "cross-env NODE_ENV=production rollup --config ../../rollup.config.js",
+    "build:dev": "cross-env rollup --config ../../rollup.config.js",
```
nilzona, line 18: no need for `cross-env` unless you're actually setting an environment variable

plugins/q/package.json
```diff
@@ -15,6 +15,7 @@
   "module": "dist/picasso-q.esm.js",
   "scripts": {
     "build": "cross-env NODE_ENV=production rollup --config ../../rollup.config.js",
+    "build:dev": "cross-env rollup --config ../../rollup.config.js",
```
nilzona, line 18: same here

veinfors:
Isn't it better to make sure the `build:watch` command builds the way you want (including source maps etc.) instead of introducing yet another build script? I feel it might cause confusion around which one to use otherwise... But maybe I'm missing something.

packages/picasso.js/package.json
quanho: Good point! Updated!

plugins/q/package.json
quanho: Updated!

plugins/hammer/package.json
quanho: Updated!

T-Wizard (approved):

package.json
```diff
@@ -7,6 +7,7 @@
   },
   "scripts": {
     "build": "cross-env NODE_ENV=production FORCE_COLOR=1 lerna run build --stream",
+    "build:dev": "FORCE_COLOR=1 lerna run build --stream",
```
T-Wizard, line 10: this is missing cross-env
```suggestion
    "build:dev": "cross-env FORCE_COLOR=1 lerna run build --stream",
```

quanho: Updated!

PR #569, opened 2021-03-18
feat: support connecting to Qlik engine to get Qlik apps

quanho:
This PR is to support connecting to Qlik engine to get data of an object from a Qlik app, instead of using a custom generated data.


studio/src/App.jsx
```diff
@@ -53,18 +73,12 @@ const App = ({ location, history }) => {
   const onCodeUpdated = React.useCallback(
     (codeData) => {
       if (selected) {
-        let { code, data } = codeData;
```
nilzona, line 56: are you sure this should be removed ?

quanho: Yes!

nilzona: but there are some handling here if any of code or data is undefined, which I believe was needed when typing in the code editor after switching examples. But I don't remember now :) ... just make sure you test properly.

quanho: The reason is that I add extra properties to codeData, e.g. dataSource (which specifies if data is from QlikEngine or not), appId, sheetId, objectId.
So to avoid many checkings like
if (typeof code === 'undefined') {
          code = selected.code;
}
I use {...selected, ...codeData}

studio/src/components/QDataArea/QDataArea.jsx
```diff
@@ -0,0 +1,409 @@
+/* eslint-disable react-hooks/exhaustive-deps */
+/* eslint-disable prefer-destructuring */
+/* eslint-disable react/prop-types */
+/* eslint-disable no-param-reassign */
+import 'regenerator-runtime/runtime'; // Polyfill for using async/await
```
veinfors, line 5: Would be better to bring this polyfill in on a higher level - `studio/src/index.js` so that async/await can be used throughout the whole project for consistency.

quanho: You can use at any level, just import it where you want to use it.

nilzona (approved):

PR #576, opened 2021-04-01
feat: add callback to decide if labels should be tilted

quanho:
This PR is to add a callback shouldAutoTilt to label to allow the user to approve if labels should be tilted or not. The problem is that picasso.js has internal function to decide if labels are tilted or not but in a number of scenarios this is not good enough, e.g. after tilting the labels take too much space comparing to the main area. In this case the user can override the picasso's decision.


cbt1:
I'm missing the added property/properties in the default setting and some describing documentation?

Also add some unit tests.

niekvanstaveren:
@quanho Should this remain open?

PR #577, opened 2021-04-06
feat: support AND/OR mode for range selections

quanho:
This PR is to allow combining range selections by using OR mode or AND mode. Currently picasso.js supports only OR mode. However, AND mode is needed for example when we want to combine two range selections, each for a measure axis in a scatter plot.

### DIFF
--- a/packages/picasso.js/src/core/component/brushing.js
+++ b/packages/picasso.js/src/core/component/brushing.js
@@ -50,7 +50,7 @@ export function styler(obj, { context, data, style, filter, mode }) {
     const len = nodes.length;
     let nodeData;
     let globalChanged = false;
-    const evaluatedDataProps = typeof dataProps === 'function' ? dataProps(brusher) : dataProps;
+    const evaluatedDataProps = typeof dataProps === 'function' ? dataProps({ brush: brusher }) : dataProps;
 
     for (let i = 0; i < len; i++) {
       // TODO - update only added and removed nodes
--- a/packages/picasso.js/src/core/component/brushing.js
+++ b/packages/picasso.js/src/core/component/brushing.js
@@ -50,6 +50,7 @@ export function styler(obj, { context, data, style, filter, mode }) {
     const len = nodes.length;
     let nodeData;
     let globalChanged = false;
+    const evaluatedDataProps = typeof dataProps === 'function' ? dataProps(brusher) : dataProps;
 
     for (let i = 0; i < len; i++) {
       // TODO - update only added and removed nodes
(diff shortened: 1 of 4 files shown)


plugins/q/src/brush/q-brush.js
```diff
@@ -285,7 +285,7 @@ export default function qBrush(brush, opts = {}, layout) {
   if (methods.rangeSelectHyperCubeValues) {
     selections.push({
       method: 'rangeSelectHyperCubeValues',
-      params: [methods.rangeSelectHyperCubeValues.path, methods.rangeSelectHyperCubeValues.ranges, [], true],
+      params: [methods.rangeSelectHyperCubeValues.path, methods.rangeSelectHyperCubeValues.ranges, [], opts.orMode ?? true],
```
cbt1, line 288: Update jsdoc with this new property

packages/picasso.js/src/core/component/brushing.js
```diff
@@ -50,6 +50,7 @@ export function styler(obj, { context, data, style, filter, mode }) {
     const len = nodes.length;
     let nodeData;
     let globalChanged = false;
+    const evaluatedDataProps = typeof dataProps === 'function' ? dataProps(brusher) : dataProps;
```
cbt1, line 53: I'm guessing that you need the brush instance to solve your particular case, but to just send that is a gonna get us in to trouble later if we suddenly need to solve some other case and need some other data. So I suggest that you at least pass in an object here instead with a `brush` property that contains the `brusher`.

quanho: brusher is an object which has a lot of properties and methods like brushes(), containsMappedData(), isActive(). I don't know if we change from brusher to {brush: brusher}, the naming is better?

plugins/q/src/brush/q-brush.js
quanho: This is not a new property. Previously, the forth parameter is always true, now we allow to set it to false.

quanho: I don't see where we address this orMode in jsdoc so I can update.

packages/picasso.js/src/core/component/brushing.js
cbt1: It's a matter of consistency. So `brush` is better than `brusher` IMO as we use `brush` else to describe the brush context. The other alternative I would think is ´{ brushContext: brusher }´

plugins/q/src/brush/q-brush.js
T-Wizard: For picasso-plugin-q this is new
and should be documented here: https://github.com/qlik-oss/picasso.js/blob/c427e83781fb667c96b356c4d1a6acee494d9fc4/plugins/q/src/brush/q-brush.js#L114-L119

packages/picasso.js/src/core/component/brushing.js
quanho: Updated!

plugins/q/src/brush/q-brush.js
quanho: Updated!

T-Wizard (approved):

PR #580, opened 2021-04-13
docs: update api

cbt1:
Attempts to fix a few errors in the spec.

I should say there are still plenty of things that are wrong or inconsistent. I'm hoping to fix most of those in smaller (easier to review) PR:s.


packages/picasso.js/src/core/chart-components/labels/labels.js
```diff
@@ -2,15 +2,20 @@ import strategies from './strategies';
 
 /**
  * @typedef {object} ComponentLabels
- * @property {string} [type='labels']
+ * @extends ComponentSettings
  */
 
 /**
+ * Component settings
  * @typedef {object} ComponentLabels.settings
```
T-Wizard, line 10:
```suggestion
 * @typedef {object} ComponentLabels~Settings
```
to be more consistent with the changes on other components

packages/picasso.js/src/core/chart-components/labels/labels.js
```diff
@@ -2,15 +2,20 @@ import strategies from './strategies';
 
 /**
  * @typedef {object} ComponentLabels
- * @property {string} [type='labels']
+ * @extends ComponentSettings
  */
```
T-Wizard, line 6: maybe an example?
(to specify the value type should have)

packages/picasso.js/src/core/chart-components/legend-seq/index.js
```diff
@@ -1,5 +1,11 @@
 import legendSeq from './legend-seq';
 
+/**
+ * @typedef {object} ComponentLegendSeq
+ * @extends ComponentSettings
+ * @property {string} scale Reference to sequential color scale
+ */
```
T-Wizard, line 7: missing?
@property {ComponentLegendSeq.setting} settings Legend settings
+
example?

packages/picasso.js/src/core/chart-components/labels/labels.js
cbt1: Yeah so I would like to use `~` only when we have multiple different settings objects. Like in the case of the axis component. If there is only a single setting it should use `.` IMO. I'll update it.

packages/picasso.js/src/core/chart-components/labels/labels.js
cbt1: All components should have examples (actually most definitions should IMO). I'm just not gonna have the time to fix that in this PR.

packages/picasso.js/src/core/chart-components/legend-seq/index.js
cbt1: `ComponentLegendSeq.settings` is definied in `legend-seq.js`?

packages/picasso.js/src/core/chart-components/labels/labels.js
cbt1: I've updated the spec to use `.` when possible and `~` only for special cases

packages/picasso.js/src/core/chart-components/legend-seq/index.js
T-Wizard: the type `ComponentLegendSeq.settings` is definied in `legend-seq.js`
but I can not find that it defines that an object of type ComponentLegendSeq has a `settings` property (of type `ComponentLegendSeq.settings`). but maybe it is there and I could just not find it.

T-Wizard:
`label` in `ComponentPoint.settings` looks like should be documented but is not
(it that intentional or you did you just miss it?)

packages/picasso.js/src/core/chart-components/property-resolver/index.js
```diff
@@ -18,26 +18,44 @@ function isPrimitive(x) {
...
+ * {
+ *  scale: '<name-of-scale>',
+ *  ref: '<data-property>'
+ * }
+ *
+ * // or explictly resolve the datum using callback function
+ * {
+ *  scale: '<name-of-scale>',
+ *  fn: (d) => Math.min(0, d.value);
+ * }
```
T-Wizard, line 44: consider removing `scale: '<name-of-scale>',`
I don't think is relevant for this example.

and I think `fn: (d) => Math.min(0, d.datum.x.value);` is a more realistic function

T-Wizard (approved):

packages/picasso.js/src/core/chart-components/legend-seq/index.js
cbt1: Yes so I believe you don't have to define the property as well

packages/picasso.js/src/core/chart-components/property-resolver/index.js
cbt1: Fixed

PR #582, opened 2021-04-15
docs: add data extraction type

cbt1:
This PR adds two new definitions for extracting data `DataExtraction` and `DataFieldExtraction`. The previously documented definition was not really reflecting reality.

### DIFF
--- a/packages/picasso.js/src/core/data/dataset.js
+++ b/packages/picasso.js/src/core/data/dataset.js
@@ -187,7 +187,7 @@ function ds({ key, data, config } = {}) {
 
     /**
      * Extract data items from this dataset
-     * @param {DataExtractConfig} config
+     * @param {DataExtractConfig|DataFieldExtraction} config
      * @returns {Array<DatumExtract>}
      */
     extract: (cfg) => extract(cfg, dataset, cache),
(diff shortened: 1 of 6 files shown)


packages/picasso.js/src/core/data/dataset.js
```diff
@@ -217,49 +217,155 @@ ds.util = {
...
- * @property {DataExtractConfig~ValueFn} [value] - The field value accessor
- * @property {DataExtractConfig~LabelFn} [label] - The field label accessor
- * @property {DataExtractConfig~TrackByFn} [trackBy] - Track by value accessor
- * @property {DataExtractConfig~ReduceFn} [reduce] - Reducer function
- * @property {DataExtractConfig~ReduceLabelFn} [reduceLabel] - Label reducer function
- * @property {DataExtractConfig~FilterFn} [filter] - Filter function
- * @property {object} [props] - Additional properties to add to the extracted item
+ * @property {DataExtraction~Extract~ValueFn} [value] - The field value accessor
+ * @property {DataExtraction~Extract~LabelFn} [label] - The field label accessor
+ * @example
```
T-Wizard, line 323: why is reduce and reduceLabel removed here as they are suppored on this level

T-Wizard (approved):

packages/picasso.js/src/core/data/dataset.js
cbt1: I'll double check if you can actually use reduce on this level as well

cbt1: Doesn't look like it

T-Wizard: there is a lot of places in la-vie that set reduceLabel on this level to avoid generating unused labels (for performace)
could it be that is it only relevant for stacked data (using picasso-plugin-q)?

cbt1: What level are you referring to? You comment is on the `props` property, is that the level your are referring to?

T-Wizard: on a property inside porps

ex:
```
props: {
  year: {
    field: 'Year',
    reduceLabel: 'none',
  },
}
```

it also reminds me that reduceLabel can be 
a function, a string refering to a common predefined function or any other string (in this case 'none') to not have any reduceLabel function (in case there are no reduceLabel it will have a default one that do client side formatting)

PR #593, opened 2021-04-28
fix: support overriding error shape for point with null size

quanho:
For a point with null size, currently we use a default error shape.
There is a need to control the size and color of the error shape.
The color of the shape should be inherited from the point (with null size).
The size should be controlled by the input.


packages/picasso.js/src/core/chart-components/point/point.js
```diff
@@ -127,15 +127,22 @@ function getType(s) {
...
-        s = DEFAULT_ERROR_SETTINGS.errorShape;
-        size = pointSize.min + s.size * (pointSize.max - pointSize.min);
+        s = {
+          fill: p.fill,
+          stroke: p.stroke,
+          strokeWidth: p.strokeWidth,
+          strokeDasharray: p.strokeDasharray,
+          opacity: p.opacity,
+          ...(errorShape || DEFAULT_ERROR_SETTINGS.errorShape),
+        };
```
johanlahti, line 144: I have seen this pattern before and don't like it since it's a risk of writing the wrong key…
Can we do something like this – or is there an ES6+ way of doing this?
```
const pickProps = (propsArr, obj) => Object.fromEntries(Object.entries(p).filter((entry) => propsArr.includes(entry[0]) ? entry : false));
s = {
  ...pickProps(['fill', 'stroke', 'strokeWidth', 'strokeDasharray', 'opacity'], p),
  ...(errorShape || DEFAULT_ERROR_SETTINGS.errorShape),
};

```

quanho: I don't want to make it complicated in this case!

johanlahti (approved):

packages/picasso.js/src/core/chart-components/point/point.js
```diff
@@ -189,7 +196,9 @@ const component = {
     const limits = extend({}, SIZE_LIMITS, this.settings.settings.sizeLimits);
     const points = resolved.items;
     const pointSize = getPointSizeLimits(resolved.settings.x, resolved.settings.y, width, height, limits);
-    return createDisplayPoints(points, this.rect, pointSize, this.settings.shapeFn || this.symbol);
+    const shapeFn = this.settings.shapeFn || this.symbol;
+    const { errorShape } = this.settings;
```
cbt1, line 200: `errorShape` is extracted from the wrong "level" of setting. It should the `const { errorShape } = this.settings.settings;`

packages/picasso.js/src/core/chart-components/point/point.js
```diff
@@ -127,15 +127,22 @@ function getType(s) {
...
       if (notNumber(size)) {
-        s = DEFAULT_ERROR_SETTINGS.errorShape;
-        size = pointSize.min + s.size * (pointSize.max - pointSize.min);
+        s = {
+          fill: p.fill,
+          stroke: p.stroke,
+          strokeWidth: p.strokeWidth,
+          strokeDasharray: p.strokeDasharray,
+          opacity: p.opacity,
+          ...(errorShape || DEFAULT_ERROR_SETTINGS.errorShape),
```
cbt1, line 143: This solution requires the `size` property to exist. Which it should not require.

cbt1:
```js
// This wont work
{
 type: 'point',
 settings: { errorShape: { shape: 'saltire' } }
}

// This do work
{
 type: 'point',
 settings: { errorShape: { shape: 'saltire', size: '10px' } }
}
```

packages/picasso.js/src/core/chart-components/point/point.js
cbt1: To clarify, `this.settings` refers to generic component settings, those that are available to all components, like `data` or `dock`

packages/picasso.js/src/core/chart-components/point/point.js
cbt1: Ideally from an API point of view, the errorShape properties should be resolved per data point. So that each data point could be configured uniquely, as all other data properties can.

packages/picasso.js/src/core/chart-components/point/point.js
```diff
@@ -1,15 +1,17 @@
...
-    stroke: '#333',
-    strokeWidth: 0,
-  },
+/**
+ * @typedef {object}
+ * @alias ComponentPoint.settings.errorShape
+ */
+const ERROR_SHAPE = {
+  /** Type of shape
+   * @type {DatumString=} */
```
cbt1, line 10: DatumString and DatumNumber means that they are resolved either as a string or number OR as a function. Which I fairly sure they are not. Just use a string and/or number.

cbt1: By function I mean a `DatumConfig` or a  `datumAccessor`

PR #597, opened 2021-05-04
feat: transformation setting for components

veinfors:
This makes it possible to trigger a "light" rendering of specific components, where only a transform is applied. The reason is to provide smoother rendering performance for charts containing components with a lot of nodes. This approach might be applicable while interacting with a chart where you want to redraw the chart continuously and maintain a responsive feeling.

Here is an example of how it can affect performance (a simple scatterplot with 2000 nodes) where chart rendering time went from 80 ms to 5 ms (the bottom one uses transfoms):
transfoms

#### Example
```js
componentDef: {
  ...,
  rendererSettings: {
    transform: () => ({ a: 1, b: 0, c: 0, d: 1, e: 100, f: 100 }),
    canvasBufferSize: { width: 1000, height: 800 },
  }
};

chart.update({ partialData: true });
```

**Checklist**

- [x] tests added
- [x] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -187,7 +187,7 @@ import componentCollectionFn from './component-collection';
 /**
  * Should return a transform object if transformation should be applied, otherwise undefined or a falsy value.
  * Transforms can be applied with the canvas, svg and dom renderer.
- * !Transform is applied when running chart.update with partialData set to true, see example.
+ * !Transform is applied when running chart.update, see example.
  * @typedef {function} RendererSettings~TransformFunction
  * @returns {TransformObject}
  * @experimental
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -187,7 +187,10 @@ import componentCollectionFn from './component-collection';
 /**
  * Should return a transform object if transformation should be applied, otherwise undefined or a falsy value.
  * Transforms can be applied with the canvas, svg and dom renderer.
- * !Transform is applied when running chart.update, see example.
+ * Transform is applied when running chart.update, see example.
+ * !Important: When a transform is applied to a component, the underlaying node representations are not updated with the new positions/sizes, which
+ * can cause problems for operations that relies on the positioning of the shapes/nodes (such as tooltips, selections etc). An extra chart update
+ * without a transform is therefore needed to make sure the node position information is in sync with the visual representation again.
  * @typedef {function} RendererSettings~TransformFunction
  * @returns {TransformObject}
  * @experimental
@@ -197,7 +200,14 @@ import componentCollectionFn from './component-collection';
  *   rendererSettings: {
  *     tranform() {
  *       if(shouldApplyTransform) {
- *         return { a: 1, b: 0, c: 0, d: 1, e: x, f: y };
+ *         return {
+ *           horizontalScaling: 1,
+ *           horizontalSkewing: 0,
+ *           verticalSkewing: 0,
+ *           verticalScaling: 1,
+ *           horizontalMoving: x,
+ *           verticalMoving: y
+ *         };
  *       }
  *     }
  *   }
@@ -217,12 +227,12 @@ import componentCollectionFn from './component-collection';
 /**
  * A format to represent a transformation.
  * @typedef {object} TransformObject
- * @property {number} a Horizontal scaling
- * @property {number} b Horizontal skewing
- * @property {number} c Vertical skewing
- * @property {number} d Vertical scaling
- * @property {number} e Horizontal moving
- * @property {number} f Vertical moving
+ * @property {number} horizontalScaling
+ * @property {number} horizontalSkewing
+ * @property {number} verticalSkewing
+ * @property {number} verticalScaling
+ * @property {number} horizontalMoving
+ * @property {number} verticalMoving
  */
 
 /**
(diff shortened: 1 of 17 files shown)


packages/picasso.js/src/web/renderer/canvas-renderer/canvas-renderer.js
```diff
@@ -160,8 +203,15 @@ export function renderer(sceneFn = sceneFactory) {
       patterns = patternizer(el.ownerDocument);
     }
 
-    const g = el.getContext('2d');
+    const g = (buffer && buffer.getContext()) || el.getContext('2d');
     const dpiRatio = dpiScale(g);
+    const transform = buffer && settings.transform();
+    if (transform) {
+      el.width = el.width; // eslint-disable-line
```
T-Wizard, line 210: why is this done? (maybe add a comment)

packages/picasso.js/src/core/chart/index.js
```diff
@@ -176,6 +177,54 @@ import componentCollectionFn from './component-collection';
...
+
+/**
+ * A format to represent a transformation.
+ * @typedef {object} TransformObject
+ * @property {number} a Horizontal scaling
+ * @property {number} b Horizontal skewing
+ * @property {number} c Vertical skewing
+ * @property {number} d Vertical scaling
+ * @property {number} e Horizontal moving
+ * @property {number} f Vertical moving
```
T-Wizard, line 225: have you considered using more descriptive names

packages/picasso.js/src/core/chart/index.js
```diff
@@ -176,6 +177,54 @@ import componentCollectionFn from './component-collection';
...
+ * @typedef {object} RendererSettings
+ * @property {RendererSettings~TransformFunction} [transform] Setting for applying transform without re-rendering the whole component completely.
+ * @property {RendererSettings~CanvasBufferSize} [canvasBufferSize] Specifies the size of buffer canvas (used together with transform setting).
+ * @experimental
+ */
+
+/**
+ * Should return a transform object if transformation should be applied, otherwise undefined or a falsy value.
+ * Transforms can be applied with the canvas, svg and dom renderer.
+ * !Transform is applied when running chart.update with partialData set to true, see example.
```
T-Wizard, line 190: it is also applied when partialData is false or missing

veinfors: Oh, you are right about that. Good catch! I haven't tested that scenario... I will do that and see if it makes sense/works in that case...

veinfors: I tested it with running just `chart.update()` as well, and that works as well. It might not be the best way to do it, but that's another story. I adjusted the description accordingly anyway!

packages/picasso.js/src/web/renderer/canvas-renderer/canvas-renderer.js
veinfors: It's to clear the canvas. I've now added a comment in the code about it.

I did read up a little on this approach of clearing the canvas, and it seems it's not the optimal way to do it... But it's done like this in other places in the canvas-renderer, so I don't want to fiddle around with that now and risk anything breaking in this PR.

It seems like a more proper and better performant way of clearing a canvas would be to use `content.clearRect(0, 0, canvas.width, canvas.height)` instead... But it could have some side effects changing to that approach straight away...

packages/picasso.js/src/core/chart/index.js
veinfors: Yeah, but it also seems to be some kind of standard around this naming (look in here for instance: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Transformations#transforms), so I feel it's ok to stick to this...

T-Wizard: the last two in not that standardized see: https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix()

packages/picasso.js/src/core/component/component-factory.js
```diff
@@ -475,6 +477,17 @@ function componentFactory(definition, context = {}) {
...
+      settings.rendererSettings &&
+      typeof settings.rendererSettings.transform === 'function' &&
+      settings.rendererSettings.transform()
+    ) {
+      rend.render();
+      return;
+    }
+
+    const nodes = (brushArgs.nodes = render.call(definitionContext, ...getRenderArgs()));
+
```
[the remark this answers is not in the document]
veinfors: You are right about that. Good catch again. I moved the code back up.

packages/picasso.js/src/core/chart/index.js
veinfors: Allright.. I've given the parameters to more descriptive names now.

packages/picasso.js/src/core/component/component-factory.js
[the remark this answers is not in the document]
veinfors: I don't think so... But actually I'm a little unsure about that, I think we might do some tuning on that in the future if we see the need for it. I just added this little note in the description of the transform feature:

"!Important: When a transform is applied to a component, the underlaying node representations are not updated with the new positions/sizes, which\ncan cause problems for operations that relies on the positioning of the shapes/nodes (such as tooltips, selections etc). An extra chart update\nwithout a transform is therefore needed to make sure the node position information is in sync with the visual representation again."

This means it's probably not suitable to use transforms as the same time as brushing (since the position of the shapes are not in sync with the visual representation).... So I think it's fine like this for now at least

T-Wizard (approved):

PR #598, opened 2021-05-07
feat: support show as a function for axis line

quanho:
This PR is to support show as a function for axis line.
This can help controlling if an axis line should be show or not based on other things. For example when the axis line and the first grid line are almost the same position, they look like a single line with larger thickness, so axis should be not shown in this case. However, when we pan or zoom the chart, e.g. a scatter plot, then the grid line is moved away from the axis line, so the axis line should be shown in this case.
TODO:
Add unit tests.


T-Wizard:
I recommend doing something like this (https://github.com/qlik-oss/picasso.js/pull/599/files) instead

PR #602, opened 2021-05-24
feat: allow to do layout before rendering

quanho:


T-Wizard:
After another look at this I think that component created when calling chart.layout are probably not mounted correctly
(missing calls to beforeMount, mount and mounted)

packages/picasso.js/src/core/chart/index.js
```diff
@@ -677,6 +677,46 @@ function chartFn(definition, context) {
...
 
+  /**
+   * Layout the chart with new settings and / or data
+   * @param {object} [def] - New chart definition
+   * @param {Array<DataSource>|DataSource} [def.data] Chart data
+   * @param {ChartSettings} [def.settings] Chart settings
+   * @param {string[]} [def.excludeFromUpdate=[]] Keys of components to not include in the update
+   * @experimental
+   */
+  instance.layout = (newProps = {}) => {
```
veinfors, line 688: How about calling the function `layoutComponents` to make it a bit more clear what it actually does, and to avoid confusion with "layout" that we often use when we talk about the data model...

packages/picasso.js/src/core/chart/index.js
```diff
@@ -677,6 +677,46 @@ function chartFn(definition, context) {
     setInteractions();
   };
 
+  /**
+   * Layout the chart with new settings and / or data
+   * @param {object} [def] - New chart definition
+   * @param {Array<DataSource>|DataSource} [def.data] Chart data
+   * @param {ChartSettings} [def.settings] Chart settings
+   * @param {string[]} [def.excludeFromUpdate=[]] Keys of components to not include in the update
```
veinfors, line 685:
```suggestion
   * @param {string[]} [def.excludeFromUpdate=[]] Keys of components to not include
```

veinfors:
Remember to update the specs as well... (by running `yarn build` in `/docs`)

packages/picasso.js/src/core/chart/index.js
quanho: Good point! Updated!

packages/picasso.js/src/core/chart/index.js
quanho: Updated!

T-Wizard:
> After another look at this I think that component created when calling chart.layout are probably not mounted correctly
> (missing calls to beforeMount, mount and mounted)
I look like it is not a problem as layoutComponent do not mark new componenets as visible and update then mounts them as they was not visible.

T-Wizard (approved):

quanho:
> Remember to update the specs as well... (by running `yarn build` in `/docs`)

Updated!

veinfors (approved):

PR #634, opened 2021-09-27
fix: support getting to be rendered nodes for animation purpose

quanho:
To support animation a component which depends on other component in term of positioning like a label depending on a point.

### DIFF
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -866,6 +866,14 @@ function chartFn(definition, context) {
     return shapes;
   };
 
+  instance.getToBeRenderedNodes = () => {
+    const shapes = [];
+    visibleComponents.forEach((c) => {
+      shapes.push(...c.instance.getToBeRenderedNodes());
+    });
+    return shapes;
+  };
+
   /**
    * Get components overlapping a point.
    * @param {Point} p - Point with x- and y-coordinate. The coordinate is relative to the browser viewport.
--- a/packages/picasso.js/src/core/component/component-factory.js
+++ b/packages/picasso.js/src/core/component/component-factory.js
@@ -596,6 +596,8 @@ function componentFactory(definition, context = {}) {
     return shapes;
   };
 
+  fn.getToBeRenderedNodes = () => brushArgs.nodes || [];
+
   fn.findShapes = (selector) => {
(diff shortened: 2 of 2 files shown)


T-Wizard:
the label compoent should probably also be updated to use this
here: https://github.com/qlik-oss/picasso.js/blob/41f0af2744628e6dbffe7613ee47601a7a3a8592/packages/picasso.js/src/core/chart-components/labels/labels.js#L55
so that is also work well with animations

quanho:
> the label compoent should probably also be updated to use this
> here:
> 
> https://github.com/qlik-oss/picasso.js/blob/41f0af2744628e6dbffe7613ee47601a7a3a8592/packages/picasso.js/src/core/chart-components/labels/labels.js#L55
> 
> 
> so that is also work well with animations

Good point! Will do that!

bao-ho:
> picasso.js/packages/picasso.js/src/core/chart-components/labels/labels.js



> the label compoent should probably also be updated to use this here:
> 
> https://github.com/qlik-oss/picasso.js/blob/41f0af2744628e6dbffe7613ee47601a7a3a8592/packages/picasso.js/src/core/chart-components/labels/labels.js#L55
> 
> 
> so that is also work well with animations

Will this require a lot of code changes in `bars`, `slices`, and `rows` strategies? Because we have to deal with two different kinds of nodes when animation is enabled vesus when it is not.

veinfors:
I have looked some more into this, and I think we should not do like this.
The only reason we have to expose "un-rendered" nodes like this is because some components are implemented to get shapes from other existing components (the label component does this), and THAT is the actual problem we should avoid.

It introduces bad dependencies and complicates the whole rendering flow a lot to make one component rely on another component. So I think we should solve this by "correcting" the implementation of such components. Like the label component, should not fetch existing points from the point component and position the labels based on that, but instead make use of what we have - the data, the scales and the component definitions to position the labels. And that applies to all components I would say - always avoid fetching shapes from existing components to make the rendering. Then the animations will work as intended!

Another drawback with this approach - getting the un-rendered nodes, like pointed out above by Bao, is that those has different properties, which would force us into handling two different node types, when animation is enabled vs disabled, which is also something we want to avoid.

With that, I'm closing this PR. If we are not able to solve it in another way, we might re-open it, but hopefully we don't need to.

PR #640, opened 2021-11-02
chore: update api compliance

joakim-ronning:
Hello!
We are migrating to a new image for doing API compliance.  
Check that the paths and names are correct to your repo.
If you have any questions don't hesitate to reach out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,7 @@ jobs:
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


veinfors:
It looks good to me (except that the api-governance job in circle is not going through :)).
I suppose this will replace the api governance check that is running from "APISpecMonitor" currently (when we trigger a release)?

PR #641, opened 2021-11-02
fix(animations): compensate for layout changes and synchronize tween phases

bao-ho:
**Checklist**

- [x] tests added
- [ ] documentation updated (not relevant)

## Support animations
- Add properties to nodes of some components so that the `trackBy` functions can grab those properties as IDs during tweening
  - Axes' tick marks and labels: add tick values
  - Grid lines: add tick values and grid lines' direction (x or y)
- Component-factory.js
  - The issue: as soon as tweening begins, the layout changes immediately to a new layout while the displaying nodes are still at the same relative positions to the old layout. This makes the nodes "jump" at the beginning of the animations.
    - To fix this, we can compesate for the layout changes by adjusting the relative positions of the displaying nodes. For example, if the new layout of the point component (the central area) jumps 10px to the left compared to the old layout, we shift the points 10px to the right, making the absolute positions of the points stay the same.
    - We call the `compensateForLayoutChanges` function of each component before doing tweening
    - A new object, `preComputedRect` is introduced to save the old layout. The new layout is in `instanceContext.rect.computed`
- Tween.js:
  - Always run all three tween phases (exit, update, and enter) even though they may be empty. This is to synchronize animations of different components, avoiding for example updating phase of one component running during exiting phase of another component.
- Unit tests: ✅

### Verification (sn scatter plot)

https://user-images.githubusercontent.com/70384379/140223607-3b27f726-beb5-4e97-85a3-aa49c9ac3f98.mov

### DIFF
--- a/packages/picasso.js/src/core/component/component-factory.js
+++ b/packages/picasso.js/src/core/component/component-factory.js
@@ -597,7 +597,7 @@ function componentFactory(definition, context = {}) {
   };
 
   fn.findShapes = (selector) => {
-    const shapes = rend.findShapes(selector);
+    const shapes = (currentTween?.inProgress() ? currentTween.targetScene : rend).findShapes(selector);
     for (let i = 0, num = shapes.length; i < num; i++) {
       appendComponentMeta(shapes[i]);
     }
--- a/packages/picasso.js/src/core/component/component-factory.js
+++ b/packages/picasso.js/src/core/component/component-factory.js
@@ -491,6 +491,9 @@ function componentFactory(definition, context = {}) {
     });
 
     if (currentNodes && settings.animations && settings.animations.enabled) {
+      /* The issue: as soon as animation begins, the layout changes immediately to a new layout while the displaying nodes' positions are still calculated relative to the old layout.
+      This makes the nodes "jump" at the beginning of the animations. To fix this, we can compesate for the layout changes by adjusting the relative positions of the displaying nodes.
+      For example, if the new layout of the point component (the central area) jumps 10px to the left compared to the old layout, we shift the points 10px to the right, making the absolute positions of the points stay the same. */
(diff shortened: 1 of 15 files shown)


packages/picasso.js/src/core/component/component-factory.js
```diff
@@ -489,6 +491,7 @@ function componentFactory(definition, context = {}) {
     });
 
     if (currentNodes && settings.animations && settings.animations.enabled) {
+      settings.animations.compensateForLayoutChanges(currentNodes, instanceContext.rect.computed, preComputedRect);
```
veinfors, line 494: First of all - don't we need to check if the `compensateForLayoutChanges` function exist before we call it here?

Then, to make it a little easier to work with, we rename the input parameters and send them in as an object, something like this:
```suggestion
      settings.animations.compensateForLayoutChanges({currentNodes, currentRect: instanceContext.rect.computed, newRect: preComputedRect});
```

packages/picasso.js/src/core/component/tween.js
```diff
@@ -58,26 +58,26 @@ function tween({ old, current }, { renderer }, config) {
           exited.ips.push(interpolateObject(ids[key], extend({}, ids[key], { r: 0.0001, opacity: 0 })));
         }
       });
-      if (exited.ips.length) {
+      if (exited.ips.length >= 0) {
```
veinfors, line 61: Is it necessary to check like this? Why if so?

bao-ho: yeah, it should be written better.

packages/picasso.js/src/core/component/component-factory.js
bao-ho: good points!

veinfors (approved):

PR #663, opened 2021-12-09
fix: support SI abbreviation only when format is auto or type is 'U'

bao-ho:
## Description (update below)
**Checklist**

- [x] tests added
- [ ] documentation updated

To fix https://jira.qlikdev.com/browse/QB-6534

Currently, any 'A' character in the format expression will force the number to adopt SI abbreviation. This seems to be an overkill. 

To give customers more freedom to use character 'A', we can restrict the SI abbreviation syntax: only when 'A' is right after format code, which uses either '#', '0', or non-breaking space. This restricted syntax is in accordance with what has been planned for the SI abbreviation syntax, see in the same file at:
https://github.com/qlik-oss/picasso.js/blob/02620ff4f8e3c89e14539bc2387c0524599322a8/plugins/q/src/formatter/parts/qs-number-formatter.js#L278

### DIFF
--- a/plugins/q/src/formatter/parts/qs-number-formatter.js
+++ b/plugins/q/src/formatter/parts/qs-number-formatter.js
@@ -514,7 +514,6 @@ class NumberFormatter {
       }
 
       value = value.replace(prep.numericRegex, (m) => {
-        console.log(m);
         if (m === t) {
           return prep.groupTemp;
         }
--- a/plugins/q/src/formatter/parts/qs-number-formatter.js
+++ b/plugins/q/src/formatter/parts/qs-number-formatter.js
@@ -145,16 +145,7 @@ function getAbbreviations(localeInfo, listSeparator) {
   return abbreviations;
 }
 
-function isSIAbbreviation(pattern, t, d) {
-  if (pattern.indexOf('A') === -1) {
-    return false;
-  }
-  const validFormatCodeCharacters = [t, d, '#', '0', '$', ' ']; // The last character is a non-breaking space, not a normal space
-  const formatCode = pattern.substring(0, pattern.indexOf('A'));
-  return [...formatCode].every((c) => validFormatCodeCharacters.includes(c));
-}
-
-function preparePattern(o, t, d) {
+function preparePattern(o, t, d, abbreviate) {
   let parts,
     lastPart,
     pattern = o.pattern,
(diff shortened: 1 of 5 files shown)


plugins/q/src/formatter/parts/qs-number-formatter.js
```diff
@@ -145,6 +145,15 @@ function getAbbreviations(localeInfo, listSeparator) {
...
 }
 
+function isSIAbbreviation(pattern) {
+  if (pattern.indexOf('A') === -1) {
+    return false;
+  }
+  const validFormatCodeCharacters = ['#', '0', ' ']; // The last character is a non-breaking space, not a normal space
+  const formatCode = pattern.substring(0, pattern.indexOf('A'));
+  return [...formatCode].every((c) => validFormatCodeCharacters.includes(c));
+}
```
T-Wizard, line 155: this need to mark all format created here: https://github.com/qlik-oss/picasso.js/blob/02620ff4f8e3c89e14539bc2387c0524599322a8/plugins/q/src/formatter/index.js#L26 as valid.
that have the format `#${decimal}##A`;

A question is if we want it to be possible for the end user to specify a format that triggers SI abbreviation
or if it should only when auto format is detected.
(if not we could replace the use of A with an extra parameter outside the pattern)

bao-ho: Thanks, @T-Wizard . I will fix the code to consider the case you mentioned.

bao-ho:
## Update

It turns out to be a better idea to disable SI abbreviation in custom format (qNumFormat.qFmt):
- The customers can use letter 'A' in a normal way (this fixes https://jira.qlikdev.com/browse/QB-6534)
- Producing same formatted strings as Engine, since Engine does not support SI abbreviation in custom format.

The reason we supported SI abbreviation with letter A is because we needed to support SI abbreviation when format is auto or type is 'U': https://github.com/qlik-oss/picasso.js/blob/02620ff4f8e3c89e14539bc2387c0524599322a8/plugins/q/src/formatter/index.js#L26
We can get rid of 'A' by simply checking if type is 'U'.

## File changed:
- plugins/q/src/data/__tests__/field.spec.js: update unit tests
- plugins/q/src/formatter/__tests__/formatter.spec.js: update unit tests (remove letter 'A' from auto format or type U patterns)
- plugins/q/src/formatter/__tests__/numberFormat.spec.js: remove unit tests that are related to custom format's SI abbreviation
- plugins/q/src/formatter/index.js: remove letter 'A' from auto format or type U pattern
- plugins/q/src/formatter/parts/qs-number-formatter.js:  instead of checking letter 'A' like before, checking type if it is 'U'.

T-Wizard (approved):

veinfors (approved):

PR #670, opened 2022-01-17
feat(picasso.js): add typescript definitions

streamside:
**Overview**

Generate typescript definitions from API specification (`scriptappy.json`). The module `@scriptappy/to-dts` is used to generate the definitions.

The JSDoc has been updated in a number of places to better describe the API. For example, when configuring picasso.js the [logging parameter](https://github.com/qlik-oss/picasso.js/pull/670/files#diff-21371e575cfa53c09eeff7afce043b4797d93f0f1a3d068970abf06c3f1d48daR92) is optional. More work can be done in this area to improve the DX.

***Steps to generate ts defs***

1. Open terminal and change directory to `<repo-root>/docs`
2. Generate new version of api specifications by running `yarn spec`
3. Generate new typescript definitions by running `yarn ts`

The generated definitions are stored in `/packages/picasso.js/types/index.d.ts`.

**Checklist**

- [x] tests added
- [x] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -337,10 +337,6 @@ export function orderComponents(element, ordered) {
 function chartFn(definition, context) {
   /**
    * @typedef {object} ChartDefinition
-<<<<<<< HEAD
-<<<<<<< HEAD
-=======
->>>>>>> 828527d (docs: remove nullable)
    * @property {ChartDefinition~beforeDestroy} [beforeDestroy]
    * @property {ChartDefinition~beforeMount} [beforeMount]
    * @property {ChartDefinition~beforeRender} [beforeRender]
@@ -349,19 +345,6 @@ function chartFn(definition, context) {
    * @property {ChartDefinition~destroyed} [destroyed]
    * @property {ChartDefinition~mounted} [mounted]
    * @property {ChartDefinition~updated} [updated]
-<<<<<<< HEAD
-=======
-   * @property {ChartDefinition~beforeDestroy?} [beforeDestroy]
-   * @property {ChartDefinition~beforeMount?} [beforeMount]
-   * @property {ChartDefinition~beforeRender?} [beforeRender]
-   * @property {ChartDefinition~beforeUpdate?} [beforeUpdate]
-   * @property {ChartDefinition~created?} [created]
-   * @property {ChartDefinition~destroyed?} [destroyed]
-   * @property {ChartDefinition~mounted?} [mounted]
-   * @property {ChartDefinition~updated?} [updated]
->>>>>>> ba6a7c6 (fix: update jsdosc)
-=======
->>>>>>> 828527d (docs: remove nullable)
    */
   let {
     /**
--- a/packages/picasso.js/src/core/chart/index.js
+++ b/packages/picasso.js/src/core/chart/index.js
@@ -107,7 +107,11 @@ import componentCollectionFn from './component-collection';
  * @property {DockLayoutSettings|customLayoutFunction} [strategy] Layout strategy used for child components.
  * @property {DataExtraction|DataFieldExtraction} [data] Extracted data that should be available to the component
  * @property {RendererSettings} [rendererSettings] Settings for the renderer used to render the component
+<<<<<<< HEAD
  * @property {string} [key] Component key
+=======
+ * @property {string} [key] TODO Add desription
+>>>>>>> ba6a7c6 (fix: update jsdosc)
  */
 
 // mark strategy as experimental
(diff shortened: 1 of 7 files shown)


veinfors:
I suspect these changes will make Api governance go read, complaining about breaking changes, but there are actually no breaking changes introduced here - only updates to the documentation...

I think we should Api governance on these changes though for verification, but I don't think we'll need to bump the major version to indicate breaking changes for this. What do you other guys think?

cbt1:
> I suspect these changes will make Api governance go read, complaining about breaking changes, but there are actually no breaking changes introduced here - only updates to the documentation...
> 
> I think we should Api governance on these changes though for verification, but I don't think we'll need to bump the major version to indicate breaking changes for this. What do you other guys think?

Are there any breaking changes here? Like actual code changes? I can only see doc change that could be breaking but once we do make a new release it shouldn't be any problem getting failed api gov checks approved for that.

packages/picasso.js/types/index.d.ts
```diff
@@ -0,0 +1,1694 @@
+declare const pjs: picassojs;
```
cbt1, line 1: It would have been nice with a disclaimer at the top of this file stating that it's autogenerated and should not be touched.

packages/picasso.js/src/core/chart/index.js
```diff
@@ -26,56 +26,52 @@ import componentCollectionFn from './component-collection';
...
+ */
+
+/**
+ * @typedef {ComponentAxis | ComponentBox | ComponentBrushArea | ComponentBrushAreaDir | ComponentBrushLasso | ComponentBrushRange | ComponentContainer | ComponentGridLine | ComponentLabels | ComponentLegendCat | ComponentLegendSeq | ComponentLine | ComponentPie | ComponentPoint | ComponentRefLine | ComponentText | ComponentTooltip} ComponentTypes
  */
 
 /**
  * @typedef {object} ChartSettings
- * @property {ComponentSettings[]} [components] Components
+ * @property {ComponentTypes[]} [components] Components
```
cbt1, line 74: What would happen if I as a user have a custom component that I'm using instead of the "native" picasso components? Would I get a type error?

cbt1:
Consider splitting it into two separate PR:s, one for the docs and one for the definition. Might be easier to reference that the "breaking" changes are only documentation changes.

packages/picasso.js/types/index.d.ts
streamside: Good idea. The file is regenerated when running `yarn ts` so it needs to be added to `@scriptappy/to-dts`. Optionally, a flag can added to the CLI options controlling if it should be included.

packages/picasso.js/src/core/chart/index.js
veinfors: That's a good point... That is something we'd like to avoid getting errors for.

streamside: It would complain that no component of the specified type exists. As a user you could either;

(1) Ignore the warning by using `// @ts-ignore`, or
(2) Extend the typescript definitions with the custom component. This could likely be done something along the lines on the guide explaining [how to extend the Material UIs theme](https://mui.com/customization/theming/#custom-variables).

streamside:
> Consider splitting it into two separate PR:s, one for the docs and one for the definition. Might be easier to reference that the "breaking" changes are only documentation changes.

Nice one. I'll move the documentation changes to a separate branch and PR.

packages/picasso.js/src/core/chart/index.js
veinfors: True, that sounds like reasonable "workaround" to me at least...

packages/picasso.js/types/index.d.ts
cbt1: Probably better to add it as an optional flag and disable it by default. To avoid introducing a change for anyone already using it.

packages/picasso.js/src/core/chart/index.js
streamside: Thought more about this and a third option is to allow any component type. It means that the IDE will supply any legal literals (e.g. axis, box, line..) but still allow the user to enter the name of a custom component. When entering the name of the custom component we should - probably - consider any properties and settings to be legal. 

The drawback is that we need to document a new component type in JSDoc. It might be that it turns up in the api docs at qlik.dev as well.

I suggest to keep the three ideas on how to solve this problem in our back pocket for now and collect some feedback about the type defs.

streamside:
> > > Consider splitting it into two separate PR:s, one for the docs and one for the definition. Might be easier to reference that the "breaking" changes are only documentation changes.
> > 
> > 
> > Nice one. I'll move the documentation changes to a separate branch and PR.
> 
> The documentation changes has been broken out to a [separate PR](https://github.com/qlik-oss/picasso.js/pull/671). It includes the JSDoc changes from this PR and `docs/scriptappy.json` is updated. I will update / clean this PR once the other one has been merged (need the updated scriptappy.json in order to generate typescript defintions).

I have updated this PR to only contain the logic related to generating type script definitions. (It shrunk to five files after the JSDoc when was extracted!)

Feel free to take a look when you have time!

veinfors (approved):

cbt1 (approved):

PR #678, opened 2022-02-07
feat: add sortNodes function in brush config

jjxie:
Add a  function called `sortNodes` in brush config.

```
{
    consume: [
      {
        data,
        context: 'selection',
        mode: 'and',
        style: {
          inactive: {},
          active: {},
        },
      },
    ],
    sortNodes: (obj) => {
      return sortedNodes;
    },
  };
```

### DIFF
--- a/packages/picasso.js/src/core/component/brushing.js
+++ b/packages/picasso.js/src/core/component/brushing.js
@@ -141,7 +141,17 @@ export function styler(obj, { context, data, style, filter, mode }) {
   const onUpdate = () => {
     const changed = update();
     if (changed) {
-      obj.renderer.render(obj.nodes);
+      if (obj.config.sortNode) {
+        const activeOpactiy = active?.opacity || 1;
+        const activeArray = obj.nodes.filter((node) => node.opacity === activeOpactiy);
+        const inactiveArray = obj.nodes.filter((node) => node.opacity !== activeOpactiy);
+        activeNodes.sort((node1, node2) => node2.r - node1.r);
+        inactiveArray.sort((node1, node2) => node2.r - node1.r);
+        const sortedNodes = inactiveArray.concat(activeArray);
+        obj.renderer.render(sortedNodes);
+      } else {
+        obj.renderer.render(obj.nodes);
+      }
     }
   };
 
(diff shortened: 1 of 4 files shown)


jjxie:
> Im guessing there is some very specific use case in mind with this change? However do consider making the API a bit more flexible, ex: by allowing the user to configure a sort callback function instead.

Yea, it is mainly to fix scatterplot point highlighting issue and don't want to effect other charts. I will check how to make it more flexible.
Screenshot 2022-02-07 at 16 33 23

cbt1:
@jjxie I see. So my point is that down the road we might need to fix some other sorting related issue for other specific case. In which case we would need to add another property/option we makes the code and the API more complex. 

So try avoid that by providing some other means of solving, ex. by using a sort callback.

packages/picasso.js/src/core/component/brushing.js
```diff
@@ -141,7 +141,11 @@ export function styler(obj, { context, data, style, filter, mode }) {
   const onUpdate = () => {
     const changed = update();
     if (changed) {
-      obj.renderer.render(obj.nodes);
+      if (typeof obj.config.sortNodes === 'function') {
+        obj.config.sortNodes(obj);
```
quanho, line 145: It should be obj.renderer.render(obj.config.sortNodes(obj));

jjxie: Changed `obj.renderer.render` back to be in Picasso

packages/picasso.js/src/core/component/brushing.js
```diff
@@ -141,7 +141,11 @@ export function styler(obj, { context, data, style, filter, mode }) {
   const onUpdate = () => {
     const changed = update();
     if (changed) {
-      obj.renderer.render(obj.nodes);
+      if (typeof obj.config.sortNodes === 'function') {
+        obj.renderer.render(obj.config.sortNodes(obj));
```
veinfors, line 145: Looks good to me. But I think we should add this `sortNodes` function in the Api spec.

veinfors (approved):

cbt1 (approved):

PR #688, opened 2022-08-08
fix: same key issue when extract data

jjxie:
Fix data not showing in two dimensional grouped bar chart when data has same qText.

The issue

Screenshot 2022-08-08 at 15 44 15

After fixing showing in nebula

Screenshot 2022-08-08 at 15 45 58

**Checklist**

- [ ] tests added
- [ ] documentation updated

### DIFF
--- a/plugins/q/src/data/extractor-t.js
+++ b/plugins/q/src/data/extractor-t.js
@@ -267,7 +267,7 @@ function getHierarchyForSMode(dataset) {
     let isNew = false;
     for (let c = 0; c < dimensions.length; c++) {
       const cell = row[dimensions[c]];
-      const key = `${id}__${cell.qText}__${cell.qElemNumber}`;
+      const key = `${id}__${cell.qElemNumber ?? cell.qElemNo}`;
       if (!keys[key]) {
         keys[key] = {
           __id: key,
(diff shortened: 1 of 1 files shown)


T-Wizard (approved):

jjxie:
Seems using `qElemNumber` is more reliable. `qNum` is optional.

plugins/q/src/data/extractor-t.js
```diff
@@ -267,7 +267,7 @@ function getHierarchyForSMode(dataset) {
     let isNew = false;
     for (let c = 0; c < dimensions.length; c++) {
       const cell = row[dimensions[c]];
-      const key = `${id}__${cell.qText}`;
+      const key = `${id}__${cell.qText}__${cell.qElemNumber}`;
```
T-Wizard, line 270: maybe
```suggestion
      const key = `${id}__${cell.qElemNumber}`;
```
using qElemNumber, qText should ned be needed.

as the function is called getHierarchyForSMode this is probably only used for cubes when qMode is 'S'
otherwise you need to use {cell.qElemNumber ?? cell.qElemNo}

jjxie: Looks like for mode s, to be safe I changed to` {cell.qElemNumber ?? cell.qElemNo}`

veinfors (approved):

PR #700, opened 2022-09-02
feat: support progressive rendering

quanho:


packages/picasso.js/src/core/chart/component-collection.js
```diff
@@ -105,15 +105,16 @@ function collectionFn({ createComponent }) {
           // Component is added
           return createComp(comp);
         }
-
+        if (comp.rendererSettings && typeof component.instance.renderer().settings === 'function') {
```
quanho, line 108: This is to fix the issue where component.instance.renderer().settings is not updated after data updated e.g. because of selection.

packages/picasso.js/src/core/component/component-factory.js
```diff
@@ -431,8 +447,23 @@ function componentFactory(definition, context = {}) {
   let currentNodes;
   let preComputedRect;
 
+  function updateBrushNodes(nodes) {
```
quanho, line 450: This is needed to support brushing otherwise brushing only hold the nodes are rendered the last time during progressive rendering.

packages/picasso.js/src/core/component/component-factory.js
```diff
@@ -431,8 +447,23 @@ function componentFactory(definition, context = {}) {
...
+    const { rendererSettings } = settings;
+    const progressive = typeof rendererSettings?.progressive === 'function' && rendererSettings.progressive();
+    if (progressive) {
+      if (progressive.start === 0) {
```
T-Wizard, line 454: why do brusing have this reset when progressive.start === 0?
there is no matching logic for the data part in component-factory.js

quanho: We need to store previously-rendered-nodes in brush during progressive-rendering.

quanho: I got your point! Thanks for that. Just updated!

packages/picasso.js/src/web/renderer/canvas-renderer/canvas-renderer.js
```diff
@@ -277,7 +280,14 @@ export function renderer(sceneFn = sceneFactory) {
...
 
     hasChangedRect = false;
-    scene = newScene;
+    if (typeof settings.progressive !== 'function' || !settings.progressive()) {
+      scene = newScene;
+      scene.shapes = shapes;
+    } else {
+      const combinedShapes = [...scene.shapes, ...shapes];
+      scene = canvasRenderer.getScene(combinedShapes);
+      scene.shapes = combinedShapes;
```
T-Wizard, line 289: Is it nesessary to recreate the scene will all the current shapes? (instead of doing it with only the new nodes and combine the results)
or is the performace cost of redoing the scene for the same not small enought to not be a problem.

veinfors: Yeah, I agree, this doesn't feel optimal...

quanho: This is safer and more correct otherwise we have to find out where children nodes to store in a scene and which can be different depending on components.

packages/picasso.js/src/core/component/component-factory.js
```diff
@@ -449,11 +449,12 @@ function componentFactory(definition, context = {}) {
...
-    if (progressive && brushArgs.nodes) {
-      brushArgs.nodes.push(...nodes);
-    } else {
+    if (typeof rendererSettings?.progressive !== 'function') {
       brushArgs.nodes = nodes;
+    } else if (rendererSettings.progressive() && brushArgs.nodes) {
+      brushArgs.nodes.push(...(nodes || []));
+    } else {
+      brushArgs.nodes = [...(nodes || [])];
     }
```
T-Wizard, line 458: why not
```
    const progressive = typeof rendererSettings?.progressive === 'function' && rendererSettings.progressive();
    if (progressive && brushArgs.nodes) {
      brushArgs.nodes = [...brushArgs.nodes, ...nodes];
    } else {
      brushArgs.nodes = nodes;
    }
```
(maybe with an extra `|| []` if needed)
I think it is safer and more understandable

quanho: I want to distinguish between progressive undefined and progressive false.

cbt1:
Docs and specs for this feature?

packages/picasso.js/src/core/component/component-factory.js
```diff
@@ -351,19 +351,32 @@ function componentFactory(definition, context = {}) {
...
         settings.data,
         { dataset: chart.dataset, collection: chart.dataCollection },
         { logger: chart.logger() },
         chart.dataCollection
       );
+      if (!progressive) {
+        data = extracted;
+      } else if (progressive.isFirst) {
+        data = extracted;
+        if (data.items) {
```
T-Wizard, line 374: why ` if (data.items) {` should it not always update data.items

quanho:
> Docs and specs for this feature?

Added!

packages/picasso.js/src/core/component/component-factory.js
T-Wizard: but don't you need to set data.items to an empty array in the case?
or how does the data.items.push(...) on the next progressive render without that.

packages/picasso.js/src/core/chart/index.js
```diff
@@ -233,6 +234,22 @@ import componentCollectionFn from './component-collection';
...
+/**
+ * A function which returns either (1) false (to specify no progressive rendering used) or an object specifing the data chunk rendered.
+ * @callback {function|undefined} RendererSettings~Progressive
+ * @returns {ProgressiveObject|boolean}
+ * @experimental
+ */
+
+/**
+ * A format to represent a data chunk to be rendered.
+ * @typedef {object} ProgressiveObject
```
cbt1, line 246: Since the consuming setting for this definition is experimental. I think this should also be `@experimental`. As I think any change to this object later will be threated as a breaking change which my not be what we want.

packages/picasso.js/src/core/chart/index.js
```diff
@@ -233,6 +234,22 @@ import componentCollectionFn from './component-collection';
  * @property {number} verticalMoving
  */
 
+/**
+ * A function which returns either (1) false (to specify no progressive rendering used) or an object specifing the data chunk rendered.
```
cbt1, line 238: Is this only for the canvas renderer? If so I think that should be stated.

packages/picasso.js/src/core/chart/index.js
quanho: Updated!

packages/picasso.js/src/core/chart/index.js
quanho: Updated!

veinfors (approved):

PR #707, opened 2022-09-15
feat: support styling brushed nodes based on active nodes

quanho:
This PR to support styling brushed nodes (e.g. border width/color) based on active nodes. E.g. in a scatter plot, the border of active nodes should be smaller when the number of active nodes are bigger.


T-Wizard:
The individual styles can already be functions we it there need for a larger change then just adding more parameters to it?

quanho:
> The individual styles can already be functions we it there need for a larger change then just adding more parameters to it? Eidt: or it that what you are doing? why lazyStyleUpdate and all the code duplication?

Updated to always have activeNodes as parameter for calculate styling

T-Wizard (approved):

PR #708, opened 2022-09-27
fix(axis): measure text correctly when checking overlapping labels

tasseKATT:


T-Wizard (approved):

packages/picasso.js/src/core/chart-components/axis/axis-label-size.js
```diff
@@ -242,7 +242,7 @@ export default function getSize({ isDiscrete, rect, formatter, measureText, scal
         !settings.labels.filterOverlapping &&
         isTiltedLabelOverlapping({
           majorTicks,
-          measureText,
+          measureText: measure,
```
cbt1, line 245: Also looks like line 225 need to change so that it uses `measure` function instead.

T-Wizard: no it only get `height` part of the measure that is the same for `measure` and `measureText`
(`measure` only clamp width to between min and max)

cbt1: I mean `const h = measureText('M').height;` should change to `const h = measure('M').height;`. Because `measureText` takes an object as argument to get the correct height.

cbt1 (approved):

packages/picasso.js/src/core/chart-components/axis/axis-label-size.js
T-Wizard: Yes. you are right. (I missed that they take different arguments)

PR #714, opened 2022-10-19
feat: add stroke and font settings to text component

PurwaShrivastava:
**Checklist**

- [ ] tests added
- [ ] documentation updated

### DIFF
--- a/packages/picasso.js/src/core/chart-components/text/text.js
+++ b/packages/picasso.js/src/core/chart-components/text/text.js
@@ -47,9 +47,12 @@ function generateTitle({ title, definitionSettings, dock, rect, measureText, sty
     dy: 0,
     anchor: getTextAnchor(dock, definitionSettings.anchor),
     baseline: 'alphabetical',
+    stroke: style.overlay.stroke || 'transparent',
+    strokeWidth: style.overlay.strokeWidth || 0,
+    fontWeight: style.overlay.fontWeight || 400 ,
   };
 
-  extend(struct, style.text);
+  extend(struct, style.text, style.overlay);
   const textRect = measureText(struct);
 
   if (dock === 'top' || dock === 'bottom') {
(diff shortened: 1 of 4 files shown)


packages/picasso.js/src/core/chart-components/text/text.js
```diff
@@ -47,9 +47,12 @@ function generateTitle({ title, definitionSettings, dock, rect, measureText, sty
     dy: 0,
     anchor: getTextAnchor(dock, definitionSettings.anchor),
     baseline: 'alphabetical',
+    stroke: style.overlay.stroke || 'transparent',
+    strokeWidth: style.overlay.strokeWidth || 0,
+    fontWeight: style.overlay.fontWeight || 400 ,
   };
 
-  extend(struct, style.text);
+  extend(struct, style.text, style.overlay);
```
T-Wizard, line 55: why have both style.text and style.overlay?
it looks like they are just merged together, why not just have all of it in style.text to begin with.
or are I missing somewhere where they are used differently?

packages/picasso.js/src/web/renderer/canvas-renderer/shapes/text.js
```diff
@@ -19,4 +19,7 @@ export default function render(t, { g, ellipsed }) {
   const bdy = baselineHeuristic(t);
 
   g.fillText(text, t.x + t.dx, t.y + t.dy + bdy);
+  if (t.stroke) {
```
T-Wizard, line 22: there is also the doStroke option in the second parameter.
but that also checks stroke-width. I am unsure if that is relevant for text.

Is there a use case for doing only strokeText and not fillText? (by setting fill to none)

packages/picasso.js/src/core/chart-components/text/text.js
PurwaShrivastava: Just wanted to keep it separate as its only used for overlay labels and not the normal text, where below I have used an existing mixin : $label-overlay

packages/picasso.js/src/web/renderer/canvas-renderer/shapes/text.js
PurwaShrivastava: I can use doStroke instead, as we have both stroke and strokeWidth for the overlayText.

no there is not that I have seen where fill is none, but strokeText is needed to use stroke and strokeWidth on text.

packages/picasso.js/src/core/chart-components/text/text.js
T-Wizard: The existing label-overlay mixin is for a totaly different use case and it may be a bad idea to reuse it for this.

my current opinion is to not have a separate overlay style

PurwaShrivastava: ok then I shall remove the mixin and use the specific styling we need inside text directly.

packages/picasso.js/src/core/chart-components/text/text.js
```diff
@@ -47,6 +47,9 @@ function generateTitle({ title, definitionSettings, dock, rect, measureText, sty
     dy: 0,
     anchor: getTextAnchor(dock, definitionSettings.anchor),
     baseline: 'alphabetical',
+    stroke: style.stroke || 'transparent',
+    strokeWidth: style.strokeWidth || 0,
+    fontWeight: style.fontWeight || 'normal',
```
T-Wizard, line 52: could this not be only
```suggestion
    stroke: 'transparent',
    strokeWidth: 0,
    fontWeight: 'normal',
```


the values from style should be handled by
```
extend(struct, style.text);
```

PurwaShrivastava: Agreed !

packages/picasso.js/src/core/chart-components/text/text.js
```diff
@@ -116,6 +119,17 @@ function generateTitle({ title, definitionSettings, dock, rect, measureText, sty
...
+
+/**
+ * @typedef {object} ComponentText.style.text
+ * @property {string} [fontSize='12px'] - Font size of text
+ * @property {string} [fontFamily='Source Sans Pro'] - Font family of text
+ * @property {string} [fontWeight='bold'] - Font weight of the overlay text
+ * @property {string} [fill='#ffffff'] - Fill color of the overlay text
+ * @property {string} [stroke='#595959'] - Stroke of the overlay text
+ * @property {number} [strokeWidth=2] - Stroke width of the overlay text
+ * @property {number} [opacity=0.5] - Opacity of the overlay text
```
tasseKATT, line 131: Should remove the word `overlay` from all of these.

PurwaShrivastava: Yes

T-Wizard (approved):

PR #729, opened 2022-11-01
fix: support adding a large number of items to an array

quanho:


packages/picasso.js/src/core/data/extractor.js
```diff
@@ -42,7 +42,12 @@ export default function extract(dataConfig, data = {}, opts = {}) {
           if (!s) {
             return;
           }
-          extracted.items.push(...s.extract(cfg));
+          const tmp = s.extract(cfg);
+          if (tmp.length < 100000) {
+            extracted.items.push(...tmp);
+          } else {
```
tasseKATT, line 48: Any purpose to keep the if-else, instead of always going with the else case?

quanho: If we push in the same array then the memory will be less than change to a new array I guess.

packages/picasso.js/src/core/data/extractor.js
```diff
@@ -42,7 +42,12 @@ export default function extract(dataConfig, data = {}, opts = {}) {
           if (!s) {
             return;
           }
-          extracted.items.push(...s.extract(cfg));
+          const tmp = s.extract(cfg);
+          if (tmp.length < 100000) {
```
cbt1, line 46: Where is the 100k value coming from?

quanho: https://stackoverflow.com/questions/14723848/push-multiple-elements-to-array

cbt1: Yes, but my point was more that it is an arbitrary and is it a magic number. We should at least create a variable with a descriptive name that holds the number and possibly add some comment as to the merit for choosing 100k.

packages/picasso.js/src/core/data/extractor.js
cbt1: Regarding my comment above, if we simple just use the spread syntax for all cases the need for that arbitrary number disappears.

It might be slower, but that is going to be a big might and may very on different browsers.

On top of that it should also be safer to just use [...ary1, ...ary2].

packages/picasso.js/src/core/data/extractor.js
cbt1: Great!

packages/picasso.js/src/core/data/extractor.js
quanho: Updated by not using push.

PR #766, opened 2023-02-26
feat: line animations

bao-ho:
**Checklist**

- [x] tests added
- [x] documentation updated

## Description

- To be able to animate lines, moving the "generating path" logic from line nodes to path display objects
- Add storage into chart, to be able to store info that is global to chart's component
More details are in the comments.

### Line animations

See https://github.com/qlik-trial/la-vie/pull/4583

### Bar animations
Making sure that the updating stage is removed properly when the current data are completely different from the old data, as before.

https://user-images.githubusercontent.com/70384379/221404775-24e45f38-8f89-481b-9531-3256bdd2d8f2.mov

### Bullet animations

Making sure that the updating stage is removed properly when the current data are completely different from the old data, as before.

https://user-images.githubusercontent.com/70384379/221418037-5d3c947f-b743-4bcd-a1df-942c43d46c74.mov

### Funnel animations

Making sure that the updating stage is removed properly when the current data are completely different from the old data, as before.

https://user-images.githubusercontent.com/70384379/221419805-274d2161-5d19-4dea-aabe-d054a449955d.mov

### Grid animations

Making sure that the updating stage is removed properly when the current data are completely different from the old data, as before.

https://user-images.githubusercontent.com/70384379/221428466-c65db9b3-c1cd-47b9-8221-2baca3b8df43.mov

### Pie animations

Making sure that the updating stage is removed properly when the current data are completely different from the old data, as before.

https://user-images.githubusercontent.com/70384379/221428991-660b4383-f0df-4eee-acc4-c626195ffe72.mov

### Sankey animations

Making sure that the updating stage is removed properly when the current data are completely different from the old data, as before.

https://user-images.githubusercontent.com/70384379/221429530-4ecb8a57-6808-43b4-8774-69eb0e6c0555.mov

### Scatter animations

Making sure that the updating stage is removed properly when the current data are completely different from the old data, as before.

https://user-images.githubusercontent.com/70384379/221564637-ff1fe8d6-c5fe-4dbe-9130-78dae463b94c.mov


packages/picasso.js/src/core/utils/object/set-value.js
```diff
@@ -0,0 +1,33 @@
...
+
+  if (value === undefined) {
+    delete scoped[propertyName];
+    return;
+  }
+
+  scoped[propertyName] = value;
+}
+
+export default setValue;
```
bao-ho, line 33: copied from 'qlik-chart-modules'

packages/picasso.js/src/core/utils/object/get-value.js
```diff
@@ -0,0 +1,28 @@
...
+      return fallback;
+    }
+
+    scoped = scoped[step];
+  }
+
+  return scoped;
+}
+
+export default getValue;
```
bao-ho, line 28: copied from 'qlik-chart-modules'

packages/picasso.js/src/core/chart/index.js
```diff
@@ -1095,6 +1096,8 @@ function chartFn(definition, context) {
 
   instance.theme = () => theme;
 
+  instance.storage = createStorage({ animations: { updatingStageMeta: { isInit: false, shouldBeRemoved: false } } });
+
```
bao-ho, line 1100: create a storage for chart. Currently, there is no place in a chart to store some info that is global to it's components. This is required in some cases. For example, when doing animations, we want to be able to remove updating stage of all components if we remove the updating stage of the main component. We want to do this in the tween function of 'tween.js', but this function belongs to each component. For each component to see the indication to remove updating stage from the main component, we need global info from the chart.

packages/picasso.js/src/core/chart-components/line/line.js
```diff
@@ -119,11 +93,15 @@ const SETTINGS = {
   },
 };
 
-function createDisplayLayer(points, { generator, item, data }, fill = '') {
-  const path = generator(points);
```
bao-ho, line 123: basically, we want to delay generating path: moving the generating path logic from line nodes to paths (display object). This is because the path above is a string and D3 interpolation, which we used to generate transitioning nodes for animations, performs poorly on graphics presented as string. Instead, we should just pass points (objects) in here, let D3 interpolates the points, and only then generate the path from these points in packages/picasso.js/src/core/scene-graph/display-objects/path.js

packages/picasso.js/src/core/chart-components/line/__tests__/line.spec.js
```diff
@@ -114,234 +111,7 @@ describe('line component', () => {
     ]);
   });
 
-  it('should render vertical line', () => {
```
bao-ho, line 117: moving all the tests for the svg strings from this file to 'path.spec.js'

packages/picasso.js/src/core/scene-graph/display-objects/__tests__/path.spec.js
```diff
@@ -30,6 +31,328 @@ describe('Path', () => {
       expect(strokes[2]).to.eql([40, 40, 0, 0, 1, -0, -40]); // arc
       expect(strokes[3]).to.eql([0, 0]); // line to
     });
+
```
bao-ho, line 34: add tests that before belong to 'line.spec.js'

packages/picasso.js/src/core/scene-graph/display-objects/path.js
```diff
@@ -42,6 +68,25 @@ export default class Path extends DisplayObject {
       arcGen.cornerRadius(v.desc.slice.cornerRadius);
       const d = arcGen(v.arcDatum);
       this.attrs.d = d;
+    } else if (v.points) {
```
bao-ho, line 71: move the generating path logic from line nodes to path display objects. See the comment in 'line.js'

packages/picasso.js/src/core/component/tween.js
```diff
@@ -37,9 +47,20 @@ export default function tween({ old, current }, { renderer }, config) {
       current.forEach((node, i) => {
         let id = trackBy(node, i);
         if (ids[id]) {
-          updated.ips.push(interpolateObject(ids[id], node));
+          if (node.type === 'path' && node.points && node.points.length > 0) {
+            const common = findCommonPointsFromTwoLines(ids[id], node);
+            updated.ips.push(
```
bao-ho, line 52: we only interpolate the common points that should be updated from the old line node to the current line node. If we don't do this and interpolate all points, the animations would look very strange and far from what we expect.

packages/picasso.js/src/core/component/tween.js
```diff
@@ -79,11 +100,15 @@ export default function tween({ old, current }, { renderer }, config) {
       if (config.isMainComponent) {
         const filterFn = config.isMainComponent?.filterFn;
         const nUpdatingNodes = filterFn ? toBeUpdated.filter(filterFn).length : toBeUpdated.length;
-        if (updatingStageMeta.isInit === false) {
-          updatingStageMeta.shouldBeRemoved = nUpdatingNodes === 0;
-          updatingStageMeta.isInit = true;
+        const { isInit, shouldBeRemoved } = chartStorage.getValue('animations.updatingStageMeta');
```
bao-ho, line 103: remove the global variable and use the chart storage instead.

T-Wizard (approved):

PR #784, opened 2023-05-09
refactor: lower device Pixel Ratio on iPhone

jjxie:
Lower the device Pixel Ratio to 2 on iPhone to reduce canvas sizes. We have some issues when a sheet has many charts, it can cause crash on iPhone due to large canvas size.
When reduce the device Pixel Ratio, the text will become blurry, but the result looks acceptable. Compare to before fix, canvas size can be reduce in half size.

Before:
IMG_0022

After
IMG_0018

Before
IMG_0025

After
IMG_0021

**Checklist**

- [ ] tests added
- [ ] documentation updated

### DIFF
--- a/packages/picasso.js/src/web/renderer/canvas-renderer/canvas-renderer.js
+++ b/packages/picasso.js/src/web/renderer/canvas-renderer/canvas-renderer.js
@@ -23,8 +23,13 @@ function toLineDash(p) {
   return [];
 }
 
+const isIPhone = typeof navigator === `undefined` ? false : /iPhone/.test(navigator.userAgent) && !window.MSStream;
+
 function dpiScale(g) {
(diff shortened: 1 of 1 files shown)


cbt1:
Does Android phones not have this issue?

jjxie:
> Does Android phones not have this issue?

The customer issue doesn't happen on Android phones, seems Android phones can handle bigger canvas size. So don't want to lower the device pixel ratio on Android. I didn't test the canvas size limitations on Android phones. We can optimize on Android phones in the future if we see similar issues.

This optimize on iPhone won't fix the crash issue, since the more charts added to the sheet, the bigger chance it exceed canvas size limits on Safari. But it can help reduce the chance exceed canvas size limits. 

A better fix probably shouldn't render charts outside of view point or paging, but it requires bigger changes as discussed. We closed specific customer bug as won't fix, but still want this optimization in.

PR #792, opened 2023-08-28
fix: show the tooltip inside the object when its takes up the whole screen

PurwaShrivastava:
This is a special case for Treemaps as there as a branch could cover the whole area available and tooltip cannot be then docked in the available positions - top, bottom, left & right.

So a new dock position is added as "ínside" where the tooltip is shown in the center of the object. Also I have removed the arrow used to dock on particular positions as we are already in the object now.

Attached is a screenshot : 

Screenshot 2023-08-28 at 09 21 33

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,7 +138,10 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/picassojs
-      - browser-tools/install-chrome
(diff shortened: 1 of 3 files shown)


T-Wizard:
add unit test for inside dock in placement.spec.js

T-Wizard (approved):

PR #839, opened 2024-03-26
fix: doesn't display all the labels properly

Donya-qlik:
**Issue:**
Some of labels are missing

**Solution:**
The labels are displayed based on a comparison between `measuredHeight` and `boundHeight` of each node, where the measured height is calculated regarding the label font style and the text + row's padding value. The result is very depends on the chart size, in smaller sizes the label might not be shown due to the `measureHeight` larger value than `boundHeight`.
In this case, there was a redundancy that resulted the row's padding value to be added twice, so for some nodes, it affected the height comparison outcome which in turn caused the label not to be shown.

An example of the result on a chart: 

**Before fixing the issue: **
Before fix

**After fixing the issue: **
image

### DIFF
--- a/packages/picasso.js/src/core/chart-components/labels/strategies/rows.js
+++ b/packages/picasso.js/src/core/chart-components/labels/strategies/rows.js
@@ -125,12 +125,11 @@ export function rows({ settings, chart, nodes, renderer, style }, placer = place
       continue;
     }
 
-    let totalHeight = 0;
+    let measuredHeight = 0;
     let measurements = [];
     let texts = [];
 
-    let maxHeight = type === 'circle' ? 2 * bounds.r * CIRCLE_FACTOR : bounds.height;
-    totalHeight += rowSettings.padding;
+    let boundHeight = type === 'circle' ? 2 * bounds.r * CIRCLE_FACTOR : bounds.height;
     let j;
     for (j = 0; j < labelSettings.length; j++) {
       let lblStngs = labelSettings[j];
@@ -140,8 +139,8 @@ export function rows({ settings, chart, nodes, renderer, style }, placer = place
       labelStruct.fontSize = `${lblStngs.fontSize}px`;
       labelStruct.text = text;
       let measured = renderer.measureText(labelStruct);
-      totalHeight += measured.height + lblStngs.padding;
-      if (totalHeight > maxHeight) {
+      measuredHeight += measured.height + lblStngs.padding;
+      if (measuredHeight > boundHeight) {
         break;
       }
       texts.push(text);
@@ -149,7 +148,7 @@ export function rows({ settings, chart, nodes, renderer, style }, placer = place
     }
 
     const labelCount = j;
-    const wiggleHeight = Math.max(0, maxHeight - totalHeight);
+    const wiggleHeight = Math.max(0, boundHeight - measuredHeight);
     let currentY;
     if (type === 'circle') {
       currentY = bounds.cy - bounds.r * CIRCLE_FACTOR;


packages/picasso.js/src/core/chart-components/labels/strategies/rows.js
```diff
@@ -125,12 +125,11 @@ export function rows({ settings, chart, nodes, renderer, style }, placer = place
...
+    let measuredHeight = 0;
     let measurements = [];
     let texts = [];
 
-    let maxHeight = type === 'circle' ? 2 * bounds.r * CIRCLE_FACTOR : bounds.height;
-    totalHeight += rowSettings.padding;
+    let boundHeight = type === 'circle' ? 2 * bounds.r * CIRCLE_FACTOR : bounds.height;
     let j;
     for (j = 0; j < labelSettings.length; j++) {
       let lblStngs = labelSettings[j];
```
Donya-qlik, line 135: [L114](https://github.com/qlik-oss/picasso.js/blob/12ac80df3a7f1c0851daa3b30c39101301169ba1/packages/picasso.js/src/core/chart-components/labels/strategies/rows.js#L114) refers to `labelSettings` which is as same as `rowSettings`. The code needs to be refactored, otherwise I think row padding should be considered once per node.

tasseKATT:
As this is a stand-alone library and open-source, the fix should be described from the libraries point of view, not from the consumers. The funnel chart should preferably not be mentioned, neither in the description of the fix nor the commit message. Not sure if should link or JIRA issues either.

cbt1:
I've looked a bit at this and concluded that your solution is not the way to solve the issue. Let me explain:

- This does not actually fix the issue. It only changes when the issues occur. Re-size the chart even smaller and it will happen again.
- Padding should be applied on all sides (left, right, top and bottom). This PR removes the bottom padding. Meaning that the label will not be correctly positioned in relation to the bottom of the "shape" it's bound to. You can even see that in your attached "fixed" screenshot.
- The reason some labels are displayed and some not, is because the shapes where the labels should fit inside does not all have the same height. Some will differ with 1 pixel or so. That is why only some labels are displayed when the height of the shape and the text + padding are almost equal.

I don't think there is a good solution to this issue but the following options are possible:
- If the goal is to try and maximize the likelyhood of labels being displayed, you could change the `padding` setting from the default value of `4` to something like `2` in the chart itself. 
- If the goal is to always show all labels or none, given a certain "funnel mode". You can look into removing the `Math.round` usage in the chart itself. That could mean that all shapes will have the same height. However that might cause other issues with floating point numbers.

Donya-qlik:
> I've looked a bit at this and concluded that your solution is not the way to solve the issue. Let me explain:
> 
> * This does not actually fix the issue. It only changes when the issues occur. Re-size the chart even smaller and it will happen again.
> * Padding should be applied on all sides (left, right, top and bottom). This PR removes the bottom padding. Meaning that the label will not be correctly positioned in relation to the bottom of the "shape" it's bound to. You can even see that in your attached "fixed" screenshot.
> * The reason some labels are displayed and some not, is because the shapes where the labels should fit inside does not all have the same height. Some will differ with 1 pixel or so. That is why only some labels are displayed when the height of the shape and the text + padding are almost equal.
> 
> I don't think there is a good solution to this issue but the following options are possible:
> 
> * If the goal is to try and maximize the likelyhood of labels being displayed, you could change the `padding` setting from the default value of `4` to something like `2` in the chart itself.
> * If the goal is to always show all labels or none, given a certain "funnel mode". You can look into removing the `Math.round` usage in the chart itself. That could mean that all shapes will have the same height. However that might cause other issues with floating point numbers.


Thanks Christofer for the explanations in details. True, I understand. As I mentioned in the description, changing the chart size might affect the results(I couldn't reproduce the same bug with my example after resizing it—until the nodes' height became very small and at that point, none of the labels were shown). This suggests that even if I completely remove the height condition (totalHeight vs maxHeight) from the original code, it doesn't guarantee that labels will always be displayed. In smaller nodes, some labels may still not be shown , so I was not sure how to think about it..

The code block where I made changes is confusing. Although I see what you mean about the positions of the central labels, I still don't fully understand how the value I removed relates to the bottom padding and why top and bottom padding are added separately when both are counted for each node. I couldn't find such a default hardcoded padding value in the funnel chart code, so I think I'll ask @niekvanstaveren for getting an idea on what the approach might be.

PR #854, opened 2024-06-04
feat(slope-ref-line): add slope to the reference line on y axis

PurwaShrivastava:
**Checklist**

- [x] tests added
- [x] documentation updated

This PR adds a slopeLine to the y-axis reference line in scatter plot.

Negative and positive slopes with label -

Screenshot 2024-07-08 at 22 32 25


veinfors:
@PurwaShrivastava  We should probably convert this pr into a draft until the implementation is more finalized, I'm thinking about support for RTL and also support for having a slope both on the X and Y axis ref lines.

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -265,6 +288,7 @@ const refLineComponent = {
           p,
           settings,
           items,
+          slopeLine,
```
quanho, line 291: We should have either p or slopeLine not both.

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -61,13 +76,15 @@ export function alignmentToNumber(align) {
  * @param {object[]} items - Array of all items (for collision detection)
  * @ignore
  */
-export function createLineWithLabel({ chart, blueprint, renderer, p, settings, items }) {
+export function createLineWithLabel({ chart, blueprint, renderer, p, settings, items, slopeLine }) {
```
quanho, line 79: p should have some more properties to cover for slope line case and we don't need to pass slopeLine here.

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -265,6 +288,7 @@ const refLineComponent = {
           p,
```
quanho, line 288: p should have more properties to cover slope line case.

quanho, line 288: We can remove p and use only slopeLine which can cover a normal ref-line.

PurwaShrivastava: Fixed

cbt1 (changes requested):
There are several lint errors that should have been detected by your code editor?

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -19,6 +19,67 @@ export function refLabelDefaultSettings() {
...
+    return scaleX((scaleY.max() - value) / slope);
+  } else if (slope < 0) {
+    return scaleX((scaleY.min() - value) / slope);
+  }
+}
+
+function getFormatter(p) {
+  let formatter;
+  if (typeof p.formatter === 'string') {
+    formatter = chart.formatter(p.formatter);
```
cbt1, line 51: Is`chart` a global variable?

cbt1: Also what is the difference between this line and line 53?

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -61,46 +122,55 @@ export function alignmentToNumber(align) {
...
-export function createLineWithLabel({ chart, blueprint, renderer, p, settings, items }) {
+export function createLineWithLabel({ chart, blueprint, renderer, p, settings, items, slopeLine }) {
   let doesNotCollide = true;
   let line = false;
+  let slope = false;
   let rect = false;
   let label = false;
   let value = false;
   let style = extend(true, {}, settings.style.line, p.line || {});
+  let slopeStyle = extend(true, refLabelDefaultSettings(), settings.style.label || {}, { fill: style.stroke });
```
cbt1, line 133: What is the difference in styling for a sloped line? Could they not share the same settings?

PurwaShrivastava:
> There are several lint errors that should have been detected by your code editor?

That's strange, let me check..

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
PurwaShrivastava: chart needs to be passed to the function, that I shall fix but the other part I have not touched / changed as it was already existing code, I have just moved it to a function so I can call it in two places.

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
PurwaShrivastava: As the label was not created in picasso before instead in scatter plot, so we had no label styling but now as we render it here we have styles sent for label from scatter plot.

PurwaShrivastava:
> > There are several lint errors that should have been detected by your code editor?
> 
> That's strange, let me check..

Fixed the lint issues

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -257,6 +268,35 @@ const refLineComponent = {
       let show = p.show === true || typeof p.show === 'undefined';
 
       if (show) {
+        let dummyLine = { ...p };
```
veinfors, line 271: I don't see the use of creating this `dummyLine` variable.

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -225,8 +323,81 @@ export function createLineWithLabel({ chart, blueprint, renderer, p, settings, i
...
+        if (slopeLine.labelStroke) {
+          rectLabel = {
+            type: 'rect',
+            x: slopeLine.isRtl ? Math.max(x, xPadding) - 2 : Math.max(x, xPadding) + 2,
+            y: Math.abs(Math.max(y, yPadding) - measured.height),
+            rx: 3,
+            ry: 3,
+            width: measured.width + slopeStyle.padding,
+            height: measured.height - 2,
+            stroke: slopeStyle.background.stroke,
```
veinfors, line 370: Do we need stroke and strokeWidth here?

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -225,8 +323,81 @@ export function createLineWithLabel({ chart, blueprint, renderer, p, settings, i
...
   // because otherwise it would collide with it's own line
-  items.push(line);
+  if (line) {
+    items.push(line);
+  } else if (slopeLine) {
+    // push slope line
+    items.push(slope);
+    // create data area labels for slope line
+    let valueString;
+    let rectLabel;
```
veinfors, line 333: Maybe rename this variable to `labelBackground` or `labelRect` to make it more clear what it is?

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -257,6 +268,35 @@ const refLineComponent = {
...
+        // Create slope line with labels
+        if (p.slope && p.slope.value !== 0) {
+          let scaleX = this.chart.scale('x');
+          let scaleY = this.chart.scale('y');
+          let minX = scaleX.min();
+          let maxX = scaleX.max();
```
veinfors, line 277: All of these should be `const` instead of `let`, right?

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -257,6 +268,35 @@ const refLineComponent = {
...
+          slopeLine = { ...dummyLine };
+          slopeLine.x1 = getPosition(scaleX, minX);
+          slopeLine.x2 = getPosition(scaleX, maxX);
+          let y1 = minX * p.slope.value + p.value;
+          let y2 = maxX * p.slope.value + p.value;
```
veinfors, line 282: Same for these - should be const instead of let

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -225,8 +323,81 @@ export function createLineWithLabel({ chart, blueprint, renderer, p, settings, i
...
+    const maxLabelWidth = 120;
+    if (slopeLine.slope.value !== 0 && (slopeLine.showLabel || slopeLine.showValue)) {
+      let slopeLabelText = slopeLine.showLabel ? slopeLine.refLineLabel : '';
+      if (slopeLine.showValue) {
+        const formatter = getFormatter(p);
```
T-Wizard, line 338:
```suggestion
        const formatter = getFormatter(p, chart);
```

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
PurwaShrivastava: ok can change it to labelRect then ...

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
PurwaShrivastava: It is defined in refLabelDefaultSettings in the file and is being used by the other rect created in the same file, so used the same as to opt for any stroke related styling ... but true not needed now... shall remove it

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -61,46 +150,55 @@ export function alignmentToNumber(align) {
...
+      strokeWidth: style.strokeWidth || 1,
+      strokeDasharray: style.strokeDasharray,
+      flipXY: p.flipXY || false, // This flips individual points (Y-lines)
+      value: p.valueInfo ? p.valueInfo.id : p.value,
+    });
+  }
 
   if (p.label) {
-    const item = extend(true, refLabelDefaultSettings(), settings.style.label || {}, { fill: style.stroke }, p.label);
+    let item = extend(true, refLabelDefaultSettings(), settings.style.label || {}, { fill: style.stroke }, p.label);
```
cbt1, line 193: Why change this to a `let` declaration? The variable does not appear to be reassigned

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -19,6 +19,95 @@ export function refLabelDefaultSettings() {
...
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+function calculateX(slopeLine, slope, maxX, measured, blueprint, slopeStyle) {
+  // calculate x for the various scenarios possible
+  let x;
```
cbt1, line 86: Looks like you can `return` instead of reassigning `x`, that should reduce the complixity a little bit

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -19,6 +19,95 @@ export function refLabelDefaultSettings() {
...
+    // TODO - Add support for array as source into formatter
+    const scaleData = p.scale.data() && p.scale.data().fields;
+    formatter = scaleData && scaleData[0] ? scaleData[0].formatter() : null;
+  }
+  return formatter;
+}
+
+function isColliding(items, slopeValue, slope, measured, maxX, xPadding, yPadding) {
+  for (let i = 0, len = items.length; i < len; i++) {
+    let curItem = items[i];
```
cbt1, line 67: Use `const`

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -19,6 +19,95 @@ export function refLabelDefaultSettings() {
...
+  // For negative slopes
+  if (slope < 0) {
+    return scaleX((scaleY.min() - value) / slope);
+  }
+  // For positive slopes
+  return scaleX((scaleY.max() - value) / slope);
+}
+
+function getFormatter(p, chart) {
+  let formatter;
```
cbt1, line 52: Use `return` instead of reassignment

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -19,6 +19,95 @@ export function refLabelDefaultSettings() {
...
+  }
+  return false;
+}
+
+function calculateX(slopeLine, slope, maxX, measured, blueprint, slopeStyle) {
+  // calculate x for the various scenarios possible
+  let x;
+  if (slopeLine.slope.value > 0) {
+    // docking at top
+    if (maxX < 0.8) {
```
cbt1, line 89: Magic number `0.8`, what is it and what is it based on? Delcare it as a `const` at the top of the file

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -182,6 +188,7 @@ const refLineComponent = {
 
   render() {
     let settings = this.settings;
+    let slopeLine;
```
cbt1, line 191: Why is `slopeLine` declared here? Can it not be declared in the scope where it's used?

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -257,6 +267,34 @@ const refLineComponent = {
       let show = p.show === true || typeof p.show === 'undefined';
 
       if (show) {
+        // Create slope line with labels
+        if (p.slope && p.slope.value !== 0) {
```
cbt1, line 271: Line 243 already checks if the value is non-zero. Is the check required here as well?

packages/picasso.js/src/core/chart-components/ref-line/index.js
```diff
@@ -14,6 +14,10 @@ import refLineComponent from './refline';
  *      label: {
  *        text: 'value label'
  *      }
+ *      slope: {
```
cbt1, line 17: What would happend if put this under the `x` line instead?

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
cbt1: I dont fully understand, could you share an example definition that sets the slope styling and the other styling?

packages/picasso.js/src/core/chart-components/ref-line/index.js
PurwaShrivastava: slope is only created for y - axis now

packages/picasso.js/src/core/chart-components/ref-line/refline.js
PurwaShrivastava: The check at line 253 checks for filtering out the lines to be drawn and the ones that are oob, but this check here is to know if we have to create a ref line or a slope line ....

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
cbt1: When looking at this line `et item = extend(true, refLabelDefaultSettings(), settings.style.label || {}, { fill: style.stroke }, p.label);` there seem to be a label styling for non-sloped line label already.

I guess the question is then, is there a difference in the API how these are set? Or are they set the same but resolved differently?

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -225,8 +324,79 @@ export function createLineWithLabel({ chart, blueprint, renderer, p, settings, i
...
+      let slopeLabelText = slopeLine.showLabel ? slopeLine.refLineLabel : '';
+      if (slopeLine.showValue) {
+        const formatter = getFormatter(p, chart);
+        const formattedValue = formatter ? formatter(slopeLine.value) : slopeLine.value;
+        valueString = `(${slopeLine.slope}x + ${formattedValue})`;
+        slopeLabelText += slopeLine.refLineLabel ? ` ${valueString}` : valueString;
+      }
+      if (slopeLabelText.length > 1) {
+        // Measure the label text
+        let measured = renderer.measureText({
```
cbt1, line 346: Use `const`

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -61,27 +151,44 @@ export function alignmentToNumber(align) {
...
-    x2: p.position,
-    y2: 1,
-    stroke: style.stroke || 'black',
-    strokeWidth: style.strokeWidth || 1,
-    strokeDasharray: style.strokeDasharray,
-    flipXY: p.flipXY || false, // This flips individual points (Y-lines)
-    value: p.valueInfo ? p.valueInfo.id : p.value,
-  });
+  if (slopeLine) {
+    slope = blueprint.processItem({
```
veinfors, line 166: I suggest you use the line variable for this as well, no need to have a specific slope variable.
That way you do not have to have the if statement further down:

```js
if (line) {
    items.push(line);
  } else if (slopeLine) {
    // push slope line
    items.push(slope);
```

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
PurwaShrivastava: p.label may be used in other charts but not in scatter plot and now with sloped lines the way they are resolved is different compared to the existing p.label..

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
veinfors: I don't see any value in differentiating the two - both are lines. It just adds yet another variable and extra conditional checks. So I think the code gets simpler by having one variable for the line no matter if it's sloped or not.

cbt1 (approved):
I have not tested this nor have I considered the API itself.

You should straive to make the code less complex and more readable, but from a code quality perspective it does match the other parts of this component which I think is fine.

packages/picasso.js/src/core/chart-components/ref-line/refline.js
```diff
@@ -90,6 +90,7 @@ function getPosition(scale, value) {
  * @property {string} [scale] - Scale to use (if undefined will use normalized value 0-1)
  * @property {ComponentRefLine~GenericObject} [line=ComponentRefLine~GenericObject] - The style of the line
  * @property {ComponentRefLine~LineLabel} [label=ComponentRefLine~LineLabel] - The label style of the line
+ * @property {number} slope - The slope for the reference line
```
veinfors, line 93:
```suggestion
 * @property {number} [slope=0] - The slope for the reference line
```

This way, we indicate that it's an optional setting, and it defaults to 0 if not specified.

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -19,6 +21,94 @@ export function refLabelDefaultSettings() {
...
+      return true;
+    }
+  } else if (slope < 0) {
+    const minY = scaleX.max() * slope + value;
+    if (minY <= scaleY.min()) {
+      return true;
+    }
+  }
+  return false;
+}
```
veinfors, line 40:
```suggestion
}

```

packages/picasso.js/src/core/chart-components/ref-line/lines-and-labels.js
```diff
@@ -225,8 +323,77 @@ export function createLineWithLabel({ chart, blueprint, renderer, p, settings, i
   // Always push the line,
   // but this is done after collision detection,
   // because otherwise it would collide with it's own line
-  items.push(line);
+  if (line) {
```
veinfors, line 326: Is it really necessary to check if line exist here?

docs/scriptappy.json
```diff
@@ -4148,6 +4148,10 @@
               "optional": true,
               "defaultValue": "ComponentRefLine~LineLabel",
               "type": "#/definitions/ComponentRefLine/definitions/LineLabel"
+            },
+            "slope": {
```
veinfors, line 4152: It looks like you haven't re-run spec generation after your last changes of the api, please do, so this get's marked as optional and get the defaultValue: 0

veinfors:
I have regression tested this PR with the chart I can think of that uses picasso ref-lines, and it works fine! 👍

I think it would be good to add some test cases for the new logic in `lines-and-labels.js` and `refline.js` before we go ahead and merge this though.

veinfors (approved):

PR #868, opened 2024-09-03
chore: update puppeteer

veinfors:


cbt1:
Are you updating the tests?

enell:
There is a newer version available that I think we should use
#863

cbt1 (approved):
Nice 🥇

PR #884, opened 2024-11-14
feat: support measuring text with font weight

quanho:
This PR to support measure text with styling e.g bold and/or italic
**Checklist**

- [ ] tests added
- [ ] documentation updated

### DIFF
--- a/packages/picasso.js/src/web/renderer/canvas-renderer/shapes/text.js
+++ b/packages/picasso.js/src/web/renderer/canvas-renderer/shapes/text.js
@@ -4,11 +4,7 @@ import { detectTextDirection, flipTextAnchor } from '../../../../core/utils/rtl-
 
 export default function render(t, { g, ellipsed, doStroke }) {
   const text = ellipsed || ellipsText(t, measureText);
-  if (t['font-weight']) {
-    g.font = `${t['font-weight']} ${t['font-size']} ${t['font-family']}`;
-  } else {
-    g.font = `${t['font-size']} ${t['font-family']}`;
-  }
+  g.font = [t['font-style'], t['font-weight'], t['font-size'], t['font-family']].filter((v) => !!v).join(' ');
   const dir = detectTextDirection(t.text);
   if (g.canvas.dir !== dir) {
     g.canvas.dir = dir;
--- a/packages/picasso.js/src/web/text-manipulation/text-metrics.js
+++ b/packages/picasso.js/src/web/text-manipulation/text-metrics.js
@@ -38,7 +38,7 @@ function setFont({ fontStyle, fontWeight, fontSize, fontFamily }) {
 }
 
 function measureTextWidth({ text, fontStyle, fontWeight, fontSize, fontFamily }) {
-  const key = text + [fontStyle, fontWeight, fontSize, fontFamily].filter((value) => !!value).join('');
+  const key = `${text} ${fontStyle} ${fontWeight} ${fontSize} ${fontFamily}`;
   if (typeof widthCache[key] !== 'number') {
     setContext();
     setFont({ fontStyle, fontWeight, fontSize, fontFamily });
--- a/packages/picasso.js/src/web/text-manipulation/text-metrics.js
+++ b/packages/picasso.js/src/web/text-manipulation/text-metrics.js
@@ -58,9 +58,9 @@ function measureTextHeight(fontSize) {
  * @private
  * @param {object} opts
  * @param {string} opts.text - Text to measure
+ * @param {string} opts.fontWeight - Font weight, e.g. 'bold'
  * @param {string} opts.fontSize - Font size with a unit definition, ex. 'px' or 'em'
  * @param {string} opts.fontFamily - Font family
- * @param {string} opts.fontWeight - Font weight, e.g. 'bold'
  * @return {object} Width and height of text in pixels
  * @example
  * measureText({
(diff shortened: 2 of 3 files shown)


johanlahti (approved):

packages/picasso.js/src/web/text-manipulation/text-metrics.js
```diff
@@ -60,8 +72,8 @@ function measureTextHeight(fontSize) {
  *  fontFamily: 'Arial'
  * }); // returns { width: 20, height: 12 }
  */
-export function measureText({ text, fontSize, fontFamily }) {
-  const w = measureTextWidth(text, fontSize, fontFamily);
+export function measureText({ text, fontSize, fontFamily, fontStyle, fontWeight }) {
+  const w = measureTextWidth({ text, fontSize, fontFamily, fontStyle, fontWeight });
```
cbt1, line 76: Have you check if `fontStyle` actually makes any difference on the measured text?

packages/picasso.js/src/web/text-manipulation/text-metrics.js
```diff
@@ -17,21 +17,31 @@ function setContext() {
...
+    isEqual(contextCache.fontStyle, fontStyle) &&
+    isEqual(contextCache.fontWeight, fontWeight) &&
+    isEqual(contextCache.fontSize, fontSize) &&
+    isEqual(contextCache.fontFamily, fontFamily)
+  ) {
     return;
   }
-
-  context.font = fontSize + ' ' + fontFamily; // eslint-disable-line
+  context.font = [fontStyle, fontWeight, fontSize, fontFamily].filter((value) => !!value).join(' '); // eslint-disable-line
```
cbt1, line 33: Is there any scenario where the output of this does not match the supported string? see https://developer.mozilla.org/en-US/docs/Web/CSS/font

packages/picasso.js/src/web/text-manipulation/text-metrics.js
```diff
@@ -17,21 +17,31 @@ function setContext() {
...
+  contextCache.fontStyle = fontStyle;
+  contextCache.fontWeight = fontWeight;
   contextCache.fontSize = fontSize;
   contextCache.fontFamily = fontFamily;
 }
 
-function measureTextWidth(text, fontSize, fontFamily) {
-  const key = text + fontSize + fontFamily;
+function measureTextWidth({ text, fontStyle, fontWeight, fontSize, fontFamily }) {
+  const key = text + [fontStyle, fontWeight, fontSize, fontFamily].filter((value) => !!value).join('');
```
cbt1, line 41: Since `measureTextWidth` is expected to be called A LOT. If I recall correctly, given a text string `Some text`, `measureTextWidth` is called for `S`, `So`, `Som`, `Some` and so forth.

So we should probably avoid running a filter function here. Since it's also executed at line 33, although not as often.

packages/picasso.js/src/web/text-manipulation/text-metrics.js
cbt1: That is what I recall as well. Unexpected, but maybe we don't need it here?

packages/picasso.js/src/web/text-manipulation/text-metrics.js
quanho: As I understand context.font has some kind of validation, so if we set something invalid, it will ignore that thing and it also does reorder a few things like putting font style before font weight (e.g. if we put font weight first and then font style in the array, the result still shows the font style first). Not sure if this helps to answer your question.

packages/picasso.js/src/web/text-manipulation/text-metrics.js
quanho: What is your solution then?

quanho: Is this better 
`[fontStyle || '', fontWeight || '', fontSize || '', fontFamily || ''].join('');`

cbt1: A string literal? Assuming that is less costly.

```js
const key = `${text} ${fontStyle} ${fontWeight} ${fontSize} ${fontFamily}`
```

packages/picasso.js/src/web/text-manipulation/text-metrics.js
cbt1: That is more that I can deduce from the documentation at MDN

cbt1: The formal syntax definition actually seem to suggest that order matters and `font-size` and `font-family` are required. So I would say that this implementation is fine given that we do expect those two attributes to be there but `font-style` and `font-weight` are optional

packages/picasso.js/src/web/text-manipulation/text-metrics.js
quanho: Removed!

packages/picasso.js/src/web/text-manipulation/text-metrics.js
quanho: Updated!

packages/picasso.js/src/web/renderer/canvas-renderer/shapes/text.js
```diff
@@ -4,11 +4,7 @@ import { detectTextDirection, flipTextAnchor } from '../../../../core/utils/rtl-
 
 export default function render(t, { g, ellipsed, doStroke }) {
   const text = ellipsed || ellipsText(t, measureText);
-  if (t['font-weight']) {
-    g.font = `${t['font-weight']} ${t['font-size']} ${t['font-family']}`;
-  } else {
-    g.font = `${t['font-size']} ${t['font-family']}`;
-  }
+  g.font = [t['font-style'], t['font-weight'], t['font-size'], t['font-family']].filter((v) => !!v).join(' ');
```
cbt1, line 7: Nitpicking here but it does not seem optional from a performance perspective to
1. Create an array
2. `filter` to create another array
3. `join` array

...everytime we render a text.

I would think [string concat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat) is a faster and provides a somewhat clean syntax. Some simpified tested in the browser console at least gave to a 50% performance increase compered to current method.

Example:
```js
g.font = (t["font-style"] ?? "").concat(t['font-weight'] ?? ", t['font-size'] ?? "", t['font-family'] ?? "")
```

PR #887, opened 2024-12-03
fix: allow padding function 2

bao-ho:
**Checklist**

- [ ] tests added
- [ ] documentation updated

### Description

This PR is a continuation from my last PR https://github.com/qlik-oss/picasso.js/pull/880
It handles the cases where paddings can be functions. This comes in handy when there is a need for flexible placement of labels that are dependent on some bars.


packages/picasso.js/src/core/chart-components/labels/strategies/bars.js
```diff
@@ -209,6 +209,10 @@ export function findBestPlacement(
 
 function approxTextBounds(label, textMetrics, rotated, rect, padding = {}) {
   const { top = PADDING, bottom = PADDING, left = PADDING, right = PADDING } = padding;
+  const leftPadding = typeof left === 'function' ? left(textMetrics) : left;
```
veinfors, line 212: This is the same padding properties as you added and documented in your previous PR, right? So there is no need for adding anything new in the api specs if I understand correct.

Could you adjust the testcase you added in. the last pr to also validate these bounds gets calculated using the functions?

quanho (approved):

PR #896, opened 2025-03-04
fix: animations improve tweening formatted label

bao-ho:
**Checklist**

- [x] tests added
- [x] documentation updated

### Description

**The issue:**

Considering animating a formatted numerical label, say from "4.5M" to "123k". The out-of-the-box d3 string interpolator produces wrong tweened results: super long strings with questionable values.

**The fix:**
  - Take advantage of the linkData properties to pass data into label nodes (the nodes are defined when creating picasso-definition for a chart)
  - Replace the wrongly interpolated text with the formatted tweened value from linkData.

### DIFF
--- a/packages/picasso.js/src/core/component/animations/adjust-tweened-nodes.js
+++ b/packages/picasso.js/src/core/component/animations/adjust-tweened-nodes.js
@@ -1,0 +1,9 @@
+export default function adjustTweenedNodes({ tweenedNodes, formatter }) {
+  tweenedNodes.forEach((node) => {
+    const { type, data } = node;
+    if (type === 'text' && data && data.value && data.formatterKey) {
+      const { value, formatterKey } = data;
+      node.text = formatter(formatterKey)(value);
+    }
+  });
+}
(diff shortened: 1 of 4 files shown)


quanho (approved):

veinfors (approved):

packages/picasso.js/src/core/component/animations/adjust-tweened-nodes.js
```diff
@@ -0,0 +1,9 @@
+export default function adjustTweenedNodes({ tweenedNodes, formatter }) {
+  tweenedNodes.forEach((node) => {
+    const { type, data } = node;
+    if (type === 'text' && data && data.value && data.formatterKey) {
```
cbt1, line 4: `formatterKey` seem to suggest that only a "key" can be used. But in reality since this just gets passed to `chart.formatter` any of the following would also be valid:

```
* instance.formatter('nameOfMyFormatter'); // Fetch an existing formatter by name
   * instance.formatter({ formatter: 'nameOfMyFormatter' }); // Fetch an existing formatter by name
   * instance.formatter({ type: 'q' }); // Fetch an existing formatter by type
   * instance.formatter({
   *  formatter: 'd3',
   *  type: 'number',
   *  format: '1.0.%'
   * }); // Create a new formatter
  ```

I suggest you change it to just `formatter` to better match reality.

bao-ho: good suggestion. I will do that. Thanks.

cbt1 (approved):

PR #897, opened 2025-03-05
feat: option to disable screenreader for axis

Moranski25:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Summary**
This PR implements the option to hide the axis-component from screen readers. If the user sets "disableScreenReader:true" in axis-settings, the axis wont be detected by a screen reader. 

disableScreenReader is not set in the axis component, so it defaults to false. The screen reader has access to the axis and reads the axis labels: 

https://github.com/user-attachments/assets/723b9faa-e57a-441e-a66d-22676dc5edb7

disableScreenReader is set to true in the axis component. The screen reader does not read the axis labels. 

https://github.com/user-attachments/assets/a342e6ba-88ef-4a9a-85cd-78937ebeddf3

### DIFF
--- a/docs/scriptappy.json
+++ b/docs/scriptappy.json
@@ -1703,6 +1703,12 @@
               "defaultValue": "auto",
               "type": "string"
             },
+            "disableScreenReader": {
+              "description": "If set to `true`, the axis will be hidden from screen readers.",
+              "optional": true,
+              "defaultValue": false,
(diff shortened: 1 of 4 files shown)


Moranski25:
It was decided to add this option to the "rendererSettings" instead, which makes this available to more components.

PR #916, opened 2025-08-06
feat: add image shape to points 

Moranski25:
**Checklist**

- [ ] tests added
- [ ] documentation updated


cbt1 (commented):
This works out-of-the-box in the svg renderer? As I'm not seeing any changes there

packages/picasso.js/src/web/renderer/canvas-renderer/canvas-renderer.js
```diff
@@ -243,15 +242,17 @@ export function renderer(sceneFn = sceneFactory) {
       buffer.apply();
       return true;
     }
+    const containsImage = Array.isArray(shapes) && shapes.some((shape) => shape.type === 'image');
```
cbt1, line 245: This seems like potentially expensive performance wise to iterate all shapes an extra time on every render. Have you noticed any difference with a large amount of shapes?

Also it's only needed if `hasChangedRect` is `true`. So should at least move it to line 252.

packages/picasso.js/src/web/renderer/canvas-renderer/shapes/image.js
```diff
@@ -0,0 +1,106 @@
...
+  };
+}
+export function positionImage(img) {
+  const position = img.imgPosition || 'center-center';
+  if (img.symbol === 'circle') {
+    const radius = Math.min(img.width, img.height) / 2;
+    img.width = radius * 2;
+    img.height = radius * 2;
+  }
+  if (position) {
```
cbt1, line 21: I would think the default switch case handles this?

packages/picasso.js/src/web/renderer/canvas-renderer/shapes/image.js
```diff
@@ -0,0 +1,106 @@
...
+
+  const ratio = window.devicePixelRatio || 1;
+  const logicalWidth = canvas.clientWidth;
+  const logicalHeight = canvas.clientHeight;
+
+  const renderWidth = logicalWidth * ratio;
+  const renderHeight = logicalHeight * ratio;
+
+  if (!offscreenBuffer || offscreenBuffer.width !== renderWidth || offscreenBuffer.height !== renderHeight) {
+    offscreenBuffer = document.createElement('canvas');
```
cbt1, line 67: Can the `canvas` element be created outside the scope of the `render` function so that all calls to render a image shares the same canvas? Seems like a small but possible optimazation

packages/picasso.js/src/core/chart-components/point/point.js
```diff
@@ -57,6 +57,27 @@ const DEFAULT_DATA_SETTINGS = {
...
+  imgScalingFactor: undefined,
+  /** Position of image
+   * @type {string=} */
+  position: 'center-center',
+  /** Size of image
+   * @type {number=} */
+  size: 0.1,
+  /** Shape of image
+   * @type {string=} */
+  symbol: 'rectangle',
```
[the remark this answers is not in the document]
Moranski25: Yes, the only shapes that it supports right now are rectangle and circle.

packages/picasso.js/src/core/scene-graph/display-objects/image.js
```diff
@@ -0,0 +1,106 @@
...
+      x: xMin,
+      y: yMin,
+      width: xMax - xMin,
+      height: yMax - yMin,
+    };
+
+    return this.__boundingRect[includeTransform];
+  }
+
+  bounds(includeTransform = false) {
```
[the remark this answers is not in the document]
Moranski25: Very true. I have fixed this in the latest update.

packages/picasso.js/src/web/renderer/canvas-renderer/canvas-renderer.js
Moranski25: I cant say that I have not noticed any difference, but I agree with the point. Since shapes is already sorted by type, I changed this to see only if the first element in shapes are of type image, instead of iterate over all the shapes.

packages/picasso.js/src/core/scene-graph/display-objects/image.js
```diff
@@ -0,0 +1,182 @@
...
+    this.__boundingRect = { true: null, false: null };
+    this.__bounds = { true: null, false: null };
+    // If no collider is provided, we add an updateCollider function that updates the collider based on the image attributes.
+    if (!v.collider) {
+      this.attrs.updateCollider = this.updateCollider.bind(this);
+    }
+  }
+
+  // This function updates the collider properties based on the image attributes. This function is called in the rendering phase.
+  updateCollider(img) {
```
cbt1, line 92: It's been 8 or 9 years since I worked on this code so my memory is failing me here but why do you need to re-create the collider during the rendering phase?

PR #990, opened 2026-05-06
fix(renderer): use CSS transform instead of top/left for sub-pixel alignment

quanho:
At non-integer device pixel ratios (e.g. 1.25x, 1.333x, 1.5x, common on Windows and high-DPI displays), positioning SVG/canvas elements with CSS `top`/`left` causes sub-pixel misalignment with their internal SVG content.

The browser layout engine snaps `top`/`left` to integer device pixel boundaries independently of SVG coordinate rendering. This means the CSS position and the SVG translate/coordinates are resolved in two separate stages with different rounding, so the final screen position of a line drawn at y=0 inside an SVG at top=2px does not necessarily equal a line drawn at y=2 inside an SVG at top=0px.

This manifests as a ~0.5-1px vertical offset between grid lines and axis tick marks, since they live in separate SVG elements positioned via `top`.

Fix: replace `left`/`top` with `transform: translate(X, Y)` on all renderer elements (svg, canvas, dom). CSS transforms go through the compositor pipeline — the same path as SVG `transform` attributes and SVG coordinates — so there is no independent device-pixel snapping and all elements share a consistent sub-pixel coordinate space.

**Checklist**

- [ ] tests added
- [ ] documentation updated

### DIFF
--- a/packages/picasso.js/src/web/renderer/dom-renderer/dom-renderer.js
+++ b/packages/picasso.js/src/web/renderer/dom-renderer/dom-renderer.js
@@ -70,8 +70,7 @@ export default function renderer(opts = {}) {
       el.setAttribute('aria-hidden', true);
     }
 
-    el.style.left = `${rect.computedPhysical.x}px`;
-    el.style.top = `${rect.computedPhysical.y}px`;
+    el.style.transform = `translate(${rect.computedPhysical.x}px, ${rect.computedPhysical.y}px)`;
     el.style.width = `${rect.computedPhysical.width}px`;
     el.style.height = `${rect.computedPhysical.height}px`;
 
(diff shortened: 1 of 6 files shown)


packages/picasso.js/src/web/renderer/dom-renderer/dom-renderer.js
```diff
@@ -70,8 +70,7 @@ export default function renderer(opts = {}) {
       el.setAttribute('aria-hidden', true);
     }
 
-    el.style.left = `${rect.computedPhysical.x}px`;
-    el.style.top = `${rect.computedPhysical.y}px`;
+    el.style.transform = `translate(${rect.computedPhysical.x}px, ${rect.computedPhysical.y}px)`;
```
T-Wizard, line 75: should the dom renderer also set left and top to '0px'?

quanho: Fixed!

T-Wizard (approved):

PR #1058, opened 2026-07-21
chore(security): add explicit minimumReleaseAge for supply-chain hardening

anupamme:
## Summary

Adds `minimumReleaseAge: 10080` to `pnpm-workspace.yaml`, requiring packages to be at least 7 days old on npm before pnpm will install them.

**Change from the original submission:** `blockExoticSubdeps` and `trustPolicy: no-downgrade` have been dropped — those need proper vetting against the dependency tree and CI/contributor workflow before landing.

## Motivation

The original PR mislabeled this as a HIGH severity vulnerability. That framing was wrong — pnpm 11.x already defaults to a 1-day release age, so there's no unprotected gap. This is a **hardening recommendation**: making the policy explicit and raising it from 1 day to 7 days to match the existing Renovate policy.

The argument for `minimumReleaseAge: 10080` specifically:
- The pre-existing `minimumReleaseAgeExclude: ["@qlik/*"]` entry already anticipates this setting — someone set up the exclude list knowing `minimumReleaseAge` would be added
- The 7-day (10080 min) value aligns direct installs with the 7-day window Renovate already enforces for automated dependency updates
- This closes the consistency gap between automated and manual install paths

## What was dropped and why

| Setting | Status | Reason |
|---|---|---|
| `minimumReleaseAge: 10080` | **Kept** | Completes existing `minimumReleaseAgeExclude` setup; aligns with Renovate policy |
| `blockExoticSubdeps: true` | Dropped | Needs dep-tree audit first |
| `trustPolicy: no-downgrade` | Dropped | Needs CI/contributor impact assessment |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -8,5 +8,7 @@ allowBuilds:
   puppeteer: false
   unrs-resolver: false
 
+minimumReleaseAge: 10080
+
 minimumReleaseAgeExclude:
   - "@qlik/*"


enell:
Thanks for the security report.

This project is using pnpm 11.15.1, where `minimumReleaseAge` already defaults to 1440 minutes (1 day). Therefore, the absence of an explicit `minimumReleaseAge` entry does not mean that newly published packages have no age restriction.

We also already configure Renovate with a 7-day `minimumReleaseAge`, which ensures that automated dependency updates are only proposed after packages have been available for that period.

The proposed pnpm setting:

```yaml
minimumReleaseAge: 10080
```

would extend the protection to direct pnpm installs as well, so it is a reasonable additional supply-chain hardening measure. However, this is a policy decision rather than a fix for an existing vulnerability.

For the other proposed settings:
- `blockExoticSubdeps: true` is a reasonable hardening option, but we would need to verify compatibility with our dependency tree.
- `trustPolicy: no-downgrade` is also worth evaluating, but we would need to confirm the impact on contributors and CI environments.

Could you clarify whether these settings are required by a specific security policy? Otherwise, I think this should be classified as a hardening recommendation rather than a high severity vulnerability.

anupamme:
Fair points, and thanks for digging into this. You're right that pnpm 11.15.1's 1-day default undercuts the "no protection at all" framing. I'll flag that as a false positive in how the scanner characterises this class of finding upstream. This isn't tied to a specific security policy on my end; it's a general supply-chain-hardening rule (recommend explicit minimumReleaseAge), not evidence of an actual exploitable gap. Agreed it should be classified as a hardening recommendation rather than a vulnerability.

That said, I think minimumReleaseAge: 10080 + the @qlik/* exclude are still worth taking, since they just extend your existing 7-day Renovate policy to direct installs; happy to narrow this PR to just those two if you'd rather evaluate blockExoticSubdeps and trustPolicy: no-downgrade separately (or drop them entirely). Let me know which you'd prefer, and I'll update the diff.

enell:
Yes please do that. The blockExoticSubdeps and-or trustPolicy configs breaks the install so we should add them separately

anupamme:
Fair points, and thanks for digging into this. You're right that pnpm 11.x's 1-day default undercuts the "no protection at all" framing — that's a false positive in how the scanner characterizes this class of finding. Agreed it should be classified as a hardening recommendation rather than a vulnerability, and the HIGH label was wrong.

This isn't tied to a specific security policy — it's a generic supply-chain-hardening rule recommending explicit `minimumReleaseAge`.

I've updated the PR to:
- **Keep** `minimumReleaseAge: 10080` — the main argument is that the repo already has `minimumReleaseAgeExclude: ["@qlik/*"]`, which anticipates this setting and would be a no-op without it. This also aligns direct installs with the 7-day window your Renovate config already enforces for automated updates.
- **Drop** `blockExoticSubdeps: true` and `trustPolicy: no-downgrade` — those need proper vetting against the dependency tree and CI/contributor workflow before landing, and I can't do that from the outside.

Happy to drop `minimumReleaseAge` too if you'd prefer to own that decision entirely, or to split it into a separate follow-up if there's a different framing that fits your contribution guidelines better.

enell (approved):

